PR #227, opened 2019-01-16
Set Swing look&feel according to the OS / graphical toolkit

pviotti:
To get from this:

screenshot from 2019-01-16 21-57-14

to this (on Linux / GTK):

screenshot from 2019-01-16 21-56-13


src/main/java/org/edumips64/Main.java
```diff
@@ -189,6 +189,10 @@ public static void main(String args[]) {
 
     showVersion();
 
+    // Set the Swing UI LookAndFeel according to the OS
+    try {
+        UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+    } catch (Exception e) {  }
```
lupino3: Please add a logging statement in the catch, instead of swallowing the exception.

lupino3:
Also, I am trying to understand why the Travis build is failing. Apparently, it cannot decrypt the encrypted variable `GH_TOKEN`, which is needed to push updates to Github Pages. (which arguably shouldn't be needed for pull requests..)

pviotti:
> 
> 
> Hmm.. For some reason, the icons with the minimized windows now overlap the existing windows. That is not desired behavior: the minimized icons should be behind the existing windows. Do you know why that is the case?

Oh, I didn't notice. I don't know, it must be some idiosyncrasy with GTK-based rendering of Swing components, because on Windows it doesn't happen.

capture

lupino3:
As discussed, the new look & feel doesn't look (& feel) great on Linux, I'm closing the pull request. Thanks anyway!

PR #234, opened 2019-01-27
Uncomment update to GitHub pages

lupino3:


lupino3:
This can't be done now, as update-gh-pages also deploys the JS code. Need to first update Gradle to compile the JS code.

PR #236, opened 2019-02-10
Converter codacity fixes

hbuie112358:
Hello. I've run the unit tests after making these changes, and all pass. I've also run the codacity tests locally, and all appear to be gone for the Converter class. I tried as much as I could not to impact performance. Please let me know if there is anything else with respect to these changes that you would like me to do. 

Thanks!

Howard


src/main/java/org/edumips64/core/Converter.java
```diff
@@ -564,6 +564,7 @@ public static long powLong(int base, long exp) {
     return ret;
   }
 
+  //Determines if a string has only 0's and 1's chars
```
lupino3: Thanks for adding comments, really appreciated! Would you mind using the javadoc conventions for them?

hbuie112358:
Hi, Andrea. Sure, I'll put the javadoc comments in. The reason I didn't was
that I was thinking you might not want private functions in javadoc.

And yes. I'll work on tests for the functions, although I think some are
already running in existing tests. I'm not at home now to verify that,
though. I'll look.

Howard

lupino3:
As far as I can see, the Converter code runs only indirectly, there are no dedicated unit tests for it.

Thanks!

hbuie112358:
Cool. I started making them this morning. Some are easier than others, will
let you know when ready. Question: I don't see where current code checks
for string length less than 64, and current code does not pass a test for
that. Would you like me to add that check in the function and as a test? I
don't think it checks that for 32 either but I'm not home now to see.

Thanks!

Howard

lupino3:
> Question: I don't see where current code checks for string length less than 64, and current code does not pass a test for that. Would you like me to add that check in the function and as a test?

What function are you referring to?

Thanks,
Andrea

hbuie112358:
Hi, Andrea. I thought I would try to finish this one before moving on to other codacity issues. I based my tests on a truth table, which is attached. I did my best not to leave out any possibilities. The attached file has the truth tables in it. I also put the comments for the new functions in javadoc format. Please have a look.

[eduMipsConverterTruthTable.xlsx](https://github.com/lupino3/edumips64/files/2873476/eduMipsConverterTruthTable.xlsx)

Question about another thing: There are codacity errors about starting functions with capital letters, like EX(). It looks like a matter of style to me. Do you want me to leave that alone? Or, if not, what should I change the names to? eX(), or ex()?

Thanks!

Howard

hbuie112358:
Sorry, closed in error.

src/main/java/org/edumips64/core/Converter.java
```diff
@@ -39,57 +39,32 @@
...
-          overflow = false;
-          break;
-        }
-      }
-
-      if (overflow) {
-        return (int)(-Math.pow(2.0, 32.0));
-      }
+    if (!unsignd && bits.length() == 32 && isOverflow(bits)) {
+      return (int)(-Math.pow(2.0, 32.0));
```
lupino3: Can you please store this value in a constant? It's more efficient than computing it every time.

Also, as a general point, if you want to do this type of operations (working with powers of two) it's better to use shift operators rather than using floating-point operations and then converting the result to int.

src/main/java/org/edumips64/core/Converter.java
```diff
@@ -119,45 +92,22 @@ public static long binToLong(String bits, boolean unsignd) throws IrregularStrin
...
-          overflow = false;
-          break;
-        }
-      }
-
-      if (overflow) {
-        return (long)(-Math.pow(2.0, 63.0));
-      }
+    if (!unsignd && bits.length() == 64 && isOverflow(bits)) {
+      return (long)(-Math.pow(2.0, 63.0));
```
lupino3: Same here, please use a constant.

src/main/java/org/edumips64/core/Converter.java
```diff
@@ -617,5 +567,70 @@ public static long powLong(int base, long exp) {
...
+   * Determines the integer value of an unsigned string of bits
+   * bits any string of only 0's and 1's
+   * @return integer value of the bit string
+   */
+  private static int getUnsignedIntValue(String bits){
+    int value = 0;
+    int i = 0;
+    for (int j = bits.length() - 1; j >= 0; j--, i++) {
+      if (bits.charAt(j) == '1') {
+        value += (int) Math.pow(2.0, (double) i);
```
lupino3: It's better to use shift operators here. Instead of

```java
value += (int) Math.pow(2.0, (double) i);
```

it's better to have
```java
value += 1 << i;
```

The left-shift operator is equivalent to 2 to the n-th power (`2 ^ i == 1 << i`), and has the advantage of not requiring conversions to floating-point and back, and also avoids any expensive operations that happen inside `Math.pow`. The left-shift operator is a single assembly instruction :)

src/main/java/org/edumips64/core/Converter.java
```diff
@@ -617,5 +567,70 @@ public static long powLong(int base, long exp) {
...
+   * Determines the long value of an unsigned string of bits
+   * bits any string of only 0's and 1's
+   * @return long value of the bit string
+   */
+  private static long getUnsignedLongValue(String bits){
+    long value = 0;
+    int i = 0;
+    for (int j = bits.length() - 1; j >= 0; j--, i++) {
+      if (bits.charAt(j) == '1') {
+        value += (long) Math.pow(2.0, (double) i);
```
lupino3: Same here, please use `1 << i`.

hbuie112358:
Hello. I've finished with the changes you requested. I am still learning git, obviously. I think I know what I did wrong, though. I'm sorry if I made things harder. Please let me know if you want me to change the branch.

Howard

src/main/java/org/edumips64/core/Converter.java
```diff
@@ -29,6 +29,9 @@
  * */
 public class Converter {
 
+  private static final int OVERFLOW_32 = -0x80000000;
```
lupino3: This should be
```java
private static final int OVERFLOW_32 = -0x10000000;
```

The number represents -2^32. Checking with Python:

```python
In [1]: hex(-2**32)
Out[1]: '-0x100000000'
```

hbuie112358: Hello. I am not very familiar with python, but it seems like maybe python's typing system doesn't constrain integers like java's does? That hex value has 9 nibbles, which is too large for java int value (I get "Integer number too large" error when I put this value in the code). And also, it seems like for this case since the msb has to be a 1, then the most significant nibble would have to be 8?

Howard

lupino3: Oh yeah, you are right, sorry about that. -2^32 cannot fit an int! You are right, thanks. My mistake was in trying to translate the old value directly, without checking the type bounds.

PR #238, opened 2019-02-16
Cleanup instruction field maint

hbuie112358:
Hi, Andrea. Here are some changes for access modifiers in the instruction set directory.

Howard


lupino3:
Looks good, thanks! Merging.

PR #242, opened 2019-02-24
Add gradle task for code coverage

pviotti:
As in https://github.com/codecov/example-gradle

Using this plugin: https://docs.gradle.org/current/userguide/jacoco_plugin.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@
 plugins {
     id 'java'
     id 'application'
+    id 'jacoco'
     id 'com.dorongold.task-tree' version '1.3.1'
 }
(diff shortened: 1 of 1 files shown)


pviotti:
BTW, code coverage reporting could also be done within Azure Pipelines, see:
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-code-coverage-results?view=azure-devops 
and 
https://docs.microsoft.com/en-us/azure/devops/pipelines/test/review-code-coverage-results?view=azure-devops
...but perhaps codecov offers more features.

Issue #222, opened 2018-11-12
Improve code quality by addressing Codacity's items


lupino3:
We enabled a static analysis tool called Codacity, which produces a report with several items that could be fixed.

The report is available here: https://app.codacy.com/project/andrea-spadaccini/edumips64/dashboard

That is a good place to start with contributions: pick some items off that list and fix them! :)

I'd like to keep this issue open as a place-holder, so don't close it but reference it from PRs. Feel free to open more specific issues if deemed useful or necessary.

pimts:
@lupino3 I'd like to help with fixing some of these Codacy issues. There seem to be quite some `Unused Imports` issues; 149 at the time of writing this comment.
Most of these `Unused Imports` come from using wildcard imports such as:
`import org.edumips64.core.*;`

What is your stance on wild card imports? Are they allowed in the project or should we explicitly import every class?

lupino3:
@pimts we didn't really think about this in the past to be honest. It was laziness on our side to use wildcard imports.

I think it's better to avoid them, so please feel free to go ahead and convert them to single-class imports.

Thank you very much for your contribution!

lupino3:
Closing this 9-year-old issue: the Codacy dashboard is now at **0 open issues** (down from ~1,700), and the configuration is set up so they stay near zero.

## What was done

**Code fixes and lint enforcement:**
- #1964 — fixed all markdown lint issues flagged by Codacy across the docs
- #1965 — markdownlint enforced in CI (`npm run lint:md`, `.markdownlint.json` — also read by Codacy)
- #1966 — the big one: fixed real Java/shell/Python/CSS issues (missing switch defaults, `new String()`, unquoted shell variables, etc.), added shellcheck to CI, and committed per-tool configs that align Codacy with the project's conventions: a curated PMD `ruleset.xml`, `.stylelintrc.json`, `.prospector.yaml`
- #1967 — default branches for the remaining defaultless switch statements
- #1968 — removed the placeholder `.eslintrc.json` after the ESLint tools were disabled on Codacy (Biome is this repo's JS linter, enforced in CI)
- #1969 — final stragglers: raw string for the LaTeX preamble (W605), `ElementTree as etree` (N817), pylint suppression in the zh Sphinx config

**Codacy configuration:**
[...]

PR #252, opened 2019-10-08
Codacity fixes

silentmatt:
More Codacity issues for #222. I left some of the instances alone where I didn't think it made a worthwhile improvement. If you disagree, I'd be happy to clean up other cases as well.


lupino3:
Thanks a lot for the improvements, @silentmatt! Looks good, merging.

Issue #199, opened 2018-07-24
Factor out the argv parsing logic


lupino3:
MainCLI currently depends on Main (eek) for command-line arguments parsing, which is itself intertwined with setting properties of the Main class.

This means the build of the CLI jar depends on the Swing sub-package, which is not great.

lupino3:
It would be nice to use a library like picocli: https://picocli.info/

iwodder:
@lupino3 if you still need/want this done, I'd be happy to pick it up. Using picocli seems straightforward.

lupino3:
Yes please @iwodder! Thanks a lot.

iwodder:
@lupino3 you're welcome. I should be able to get something up in the next few days. Any preferences on the picocli version? I was going to use their latest which is 4.5.0.

lupino3:
The latest stable version would be perfect. Thanks!

lupino3:
This was fixed by @iwodder in #448. Thanks!

PR #253, opened 2019-10-13
Factor out argv parsing

silentmatt:
See #199

This doesn't address the dependency on Main in the build file, since I'm not familiar enough with Gradle to confidently make the change there. It should remove the dependency from Java's standpoint though. I named the class ApplicationSettings since I had to move a few other fields (VERSION, BUILD_DATE, etc.) into the class, so it's a little more than just argument parsing.


lupino3:
Closing this as it was fixed by another contributor with updates to `MainCLI` as well.

Issue #257, opened 2020-01-14
Improve handling of out-of-memory issues


lupino3:
- [x] improve the error message to make the user understand what's going on
- [x] only report the first out-of-memory error in the Parser
- [ ] add and explain data limit to the manual

PR #266, opened 2020-01-20
OOM-related fixes

lupino3:


lupino3:
Addresses the first 2 items in  #257 .

Issue #260, opened 2020-01-17
Make DATA_LIMIT customizable via UI


lupino3:
After the investigation in #259, if it makes sense we should let users customize `DATA_LIMIT` from the UI.

This will also require setting an upper bound and fixing the documentation.

lupino3:
I thought about this, and there is no reason to make this customizable.

In the 10+ years of the simulator being around, this has proven to be an issue only once, and even if we made it customizable we'd have to choose a reasonable upper bound, to avoid OOMing.

I'll bump up memory to 640 kB. That ought to be enough for anybody :) (should be a reasonable limit for educational programs). If more people complain about this limit, I'll do the work to make it customizable, but otherwise there are more important things to do.

PR #268, opened 2020-01-21
Bump up data memory to 640 kB

lupino3:
That ought to be enough for everybody.

Also bump up code memory.


lupino3:
This addresses the memory increase discussed in #260 .

PR #309, opened 2020-04-12
Augment version number for CI builds

lupino3:
This will help distinguish JARs coming from CI.


lupino3:
The logic works well, but both the commit ID and the branch are bogus under GitHub actions: need to use the appropriate env variables: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables (`GITHUB_SHA` and `GITHUB_REF`)

PR #312, opened 2020-04-12
Try and fix packaging issues

### LINKED ISSUE #311: Issues with v1.2.7 packages
There are some packaging issues with the freshly released v1.2.7:
 - the slim jar does not contain the docs folder (which is required to get the help manual)
 - images, resources and compiled class files are copied twice in the jar (open one of the jar files to see this)

I think I've fixed this, I'll submit a PR.

pviotti:
- updates `task-tree` task (because it wasn't working)
 - move `copyHelp` as dependency of `jar` task instead of `standaloneJar` so that the "slim" jar contains the `docs` folder too
 - comment out the GWT plugin and its tasks. It turns out the double copy issue is due to the fact that the GWT plugin for some reason puts [`processResources` as dependency of `compileJava`](https://github.com/ascendtech/gwt-gradle/blob/778ea93cdde928e8b09ff26c23ef45ba86af0058/src/main/groovy/us/ascendtech/gwt/lib/GWTLibPlugin.groovy#L107), so you end up with `processResources` executed twice ([ref](https://docs.gradle.org/current/userguide/java_plugin.html#java_plugin)). This is clearly not ideal, but I'm not sure what else can be done aside from commenting it out (and maybe opening an issue in the repo of that plugin) - suggestions are welcome.

image

Close #311


lupino3:
Thanks for investigating this! Seems like they have an open bug about this: https://github.com/ascendtech/gwt-gradle/issues/3. I'll see if 0.4.23 indeed fixes the bug and report back. In the same issue they also describe an option for the `jar` task to deal with duplicates, which we could use if all else fails.

lupino3:
@pviotti 0.4.23 doesn't seem to suffer from the same issue. I tried building locally and unzipping the JAR does not give me any more warnings about duplicate files.

Could you please bump `us.ascendtech.gwt.classic` up to `0.4.23` and re-enable it? Thanks!

pviotti:
I've bumped it to the most recent version, and it seems to be ok, though the tasks duplication is still there: so I'm not sure what was wrong. I've also updated gradle wrapper to the most recent version. (I also suggest adopting one of those update dependencies bots like [dependabot](https://dependabot.com/java/) or greekeeper to take care of this kind of chores). HTH

lupino3:
Filed #314 to keep track of your suggestion, thanks a lot!

PR #322, opened 2020-04-16
snap: add desktop and menu integration

lupino3:
Work for #8


lupino3:
I think there was a race condition somewhere: this PR got merged into master but it's marked as open. I'll close it.

PR #369, opened 2020-05-15
Use GWT inside web worker

lupino3:


src/main/java/org/edumips64/webclient.gwt.xml
```diff
@@ -2,6 +2,14 @@
   <inherits name="com.google.gwt.user.User"/>
```
niloc132, line 2: User makes almost no sense in a web worker, I advise turning this down to just Core (and also removing at least Logging, since that requires widgets too i believe - JSON ... there is probably a nicer way to do that, i dont know what it requires off the top of my head).

src/main/java/org/edumips64/webclient.gwt.xml
```diff
@@ -2,6 +2,14 @@
...
   <inherits name="com.google.gwt.logging.Logging"/>
   <inherits name="com.google.gwt.json.JSON" />
+
+	<inherits name="org.gwtproject.rpc.worker.RpcToWorkers" />
+
+  <inherits name="prj.gwtwwlinker.workerlinker.WorkerLinker" />
+  <!-- Settings required for Web Worker support. A single permutation must be compiled. -->
+  <set-property name="user.agent" value="safari" />
+  <set-configuration-property name="user.agent.runtimeWarning" value="false" />
```
niloc132, line 11: this is the right idea to disable the checks, but this is the user agent check, not the compatMode check, you want both turned off. If you are able to simplify your .gwt.xml to remove the other modules though, these are unnecessary (and will cause errors if they are present).

src/main/java/org/edumips64/webclient.gwt.xml
lupino3: Thanks for the suggestion, removed!

This also removed the need for the `user.agent` property to be changed.

src/main/java/org/edumips64/webclient.gwt.xml
lupino3: Yes, thanks a lot! It seems to compile correctly now. Still not working fully but I have something to work with now!

src/main/java/org/edumips64/client/Worker.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.google.gwt.core.client.EntryPoint;
+import org.gwtproject.rpc.worker.client.worker.MessagePort;
+import org.gwtproject.rpc.worker.client.worker.MessageEvent;
+
+import jsinterop.annotations.JsType;
+
+import java.util.logging.Logger;
+
+@JsType(namespace = "jsedumips64")
```
niloc132, line 33: Unless you are running other JS inside this worker, this annotation is unnecessary (and may increase code size, possibly cause extra compile warnings).

src/main/java/org/edumips64/client/Worker.java
```diff
@@ -0,0 +1,70 @@
...
+    postMessage("ready");
+  }
+  
+  private final void route(MessageEvent event) {
+    info("GOT A MESSAGE");
+    info(event.getData());
+    // TODO: route calls to Simulator and send back messages.
+  }
+
+  private native MessagePort self() /*-{
```
niloc132, line 55: This should not be needed now that you fixed that window=$wnd issue - you can probably instead just do jsinterop instead. DomGlobal.postMessage or DomGlobal.window.postMessage, DomGlobal.window.addEventListener("message", ...), etc

src/main/java/org/edumips64/client/Worker.java
lupino3: Done, thanks!

src/main/java/org/edumips64/client/Worker.java
lupino3: Makes sense, thanks! Done.

Issue #344, opened 2020-04-25
Web UI: basic memory widget


lupino3:

lupino3:
Tracked in #852.

Issue #345, opened 2020-04-25
Web UI: display code information (was: implement the Code Java widget)


lupino3:

lupino3:
It is probably worth not having a separate widget, but instead displaying the information in the editor in some way (e.g., with a hover: https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.hover.html)

lupino3:
And use something like glyphs to mark which instruction is in a given stage: https://microsoft.github.io/monaco-editor/playground.html#interacting-with-the-editor-rendering-glyphs-in-the-margin

PR #401, opened 2020-05-30
Minor style fixes for memory textarea

pviotti:
A couple of style fixes to the memory textarea (I know there are better plans for that as per #344, but in the meantime...): 
- invert the CSS order so that `width:100% height:100%` in style.css is applied to the textarea element (now it's rendered shrunk)
- use same text color as in the other elements, which is set by materialize.css (now it's black)

---

Unrelated to this PR, some feedback on the Web UI:
 - it's pretty cool, good job! :+1: 
 - the Monaco editor is quite heavy: it makes the js bundle about 2.7MB, from just 300kB with a simple textarea instead... I wonder if there's any way to make it slimmer (or use Ace or something [else](https://en.wikipedia.org/wiki/Comparison_of_JavaScript-based_source_code_editors)) :thinking: 
 - etc (other stuff, but no need to be too picky on a prototype :smiley: )

### DIFF
--- a/src/webapp/static/index.html
+++ b/src/webapp/static/index.html
@@ -2,13 +2,14 @@
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8">
     <title>EduMIPS64 Experimental Web Frontend</title>
-      <link rel="stylesheet" media="screen" type="text/css" href="style.css">
       <link rel="icon" type="image/png" href="favicon.png">
 
        <!-- Materialize.CSS -->
       <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
-      <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+      <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" >
       <script src="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/js/materialize.min.js"></script>
+
+      <link rel="stylesheet" media="screen" type="text/css" href="style.css">
 
       <!-- Materialize JS initialization code -->
       <script type="text/javascript" language="javascript">
--- a/src/webapp/static/style.css
+++ b/src/webapp/static/style.css
@@ -33,6 +33,7 @@ textarea {
     width: 100%;
     height: 100%;
     border: 0;
+    color: rgba(0,0,0,0.87)
 }
 
 /* Registers table */


lupino3:
> A couple of style fixes to the memory textarea (I know there are better plans for that as per #344, but in the meantime...):
> 
> * invert the CSS order so that `width:100% height:100%` in style.css is applied to the textarea element (now it's rendered shrunk)
> * use same text color as in the other elements, which is set by materialize.css (now it's black)

Thanks a lot for the fix!

> Unrelated to this PR, some feedback on the Web UI:
> 
> * it's pretty cool, good job! +1

Thanks! It's very slowly starting to take shape. Still lots to do. I welcome any and all feedback!

> * the Monaco editor is quite heavy: it makes the js bundle about 2.7MB, from just 300kB with a simple textarea instead... I wonder if there's any way to make it slimmer (or use Ace or something [else](https://en.wikipedia.org/wiki/Comparison_of_JavaScript-based_source_code_editors)) thinking

Yes, it is pretty heavy. I think we can include only a subset of functions with the webpack plugin: https://github.com/microsoft/monaco-editor-webpack-plugin (look for "features").

> * etc (other stuff, but no need to be too picky on a prototype smiley )

There is still lots of work to do, it's a matter of time and priority, but if you have more feedback please do write it down so it can inform the development and the priorities! :)

I'm learning React and modern JS as the project progresses, so any feedback is welcome. My last web design work happened ~10 years ago XD

Here are the few things I have in mind that I haven't captured in issues:

- the statistics window should go away, replaced by a steps counter that, when clicked, shows the rest of statistics;
- the pipeline widget could be collapsed to a small widget that shows which stages are full/empty, and on hover show instruction info;
- this will allow to have fewer widgets; the main UI should then be prominently the editor, registers and the memory (code will probably be incorporated in the editor as per #345.

pviotti:
> There is still lots of work to do, it's a matter of time and priority, but if you have more feedback please do write it down so it can inform the development and the priorities! :)

Nothing special really, and I'm not expert either. Maybe eventually I'd consider
 - using a ui component library (like [material-ui](https://material-ui.com/) or [fluent-ui](https://developer.microsoft.com/en-us/fluentui/#/controls/web)) instead of having the Materialize css disjoint from React
 - modeling state with some more sophisticated framework like Redux, or some FSM library (e.g. [xstate](https://github.com/davidkpiano/xstate/tree/master/packages/xstate-fsm) - never used it though)
 - using Typescript - in my experience it gets handy to have stronger typing when things get more complex

but some of this would introduce additional complexity that would be hardly justifiable at this stage (actually, if I ever learned anything in my limited experience with React/nodejs is that you should minimize complexity and useless dependencies... it gets untenable very quickly).

lupino3:
> > There is still lots of work to do, it's a matter of time and priority, but if you have more feedback please do write it down so it can inform the development and the priorities! :)
> 
> Nothing special really, and I'm not expert either. Maybe eventually I'd consider
> 
> * using a ui component library (like [material-ui](https://material-ui.com/) or [fluent-ui](https://developer.microsoft.com/en-us/fluentui/#/controls/web)) instead of having the Materialize css disjoint from React

Yup, absolutely. I've played a bit with it already in a branch, but it's far from publishable. This is actually mentioned here: (https://github.com/EduMIPS64/edumips64/blob/master/docs/design/web-ui-roadmap.md#development-features `D.2`)

I was playing with materialize.css back when I didn't have the JS toolchain set up and was loading everything from CDNs; now it's much easier to use external libraries.

> * modeling state with some more sophisticated framework like Redux, or some FSM library (e.g. [xstate](https://github.com/davidkpiano/xstate/tree/master/packages/xstate-fsm) - never used it though)

Yeah I have absolutely 0 experience here, not sure if this would help or not. The state management in React so far is not very complex, maybe we should investigate if the complexity becomes difficult to manage.

> * using Typescript - in my experience it gets handy to have stronger typing when things get more complex

Agreed! I'm already re-learning Javascript though 🤣  I think it makes sense to investigate this kind of improvement after an MVP is out, assuming this wouldn't make it the development much easier (I think for now vanilla JS is fine)

> but some of this would introduce additional complexity that would be hardly justifiable at this stage (actually, if I ever learned anything in my limited experience with React/nodejs is that you should minimize complexity and useless dependencies... it gets untenable very quickly).

Thanks for the suggestion! Will keep it in mind :)

Any contribution (including this kind of edifying discussion) is more than welcome :wink: thank you!

PR #402, opened 2020-05-30
Use webpack plugins to copy files and add webpack-dev-server

pviotti:
- replace "manual" copy of static files to build dir with a webpack plugin
 - replace serve with [webpack-dev-server](https://webpack.js.org/guides/development/#using-webpack-dev-server) (which provides live code reloading)
 - update pipelines accordingly (remove "deploy" step)
 - update dev docs
 - remove replaced dependencies


lupino3:
Thank you!

PR #439, opened 2020-06-22
Fixes a "Use Equals To Compare String"

pimts:
Fixes a "Use Equals To Compare String" item from the Codacity issues.
See https://github.com/EduMIPS64/edumips64/issues/222


lupino3:
Thanks for this fix!

PR #440, opened 2020-06-24
Fixes "Unused Imports" issues from the Codacy.

pimts:
Fixes "Unused Imports" issues from the Codacy.

See https://github.com/EduMIPS64/edumips64/issues/222

### DIFF
--- a/src/main/java/org/edumips64/core/is/FPArithmeticInstructions.java
+++ b/src/main/java/org/edumips64/core/is/FPArithmeticInstructions.java
@@ -24,10 +24,19 @@
 
 package org.edumips64.core.is;
 
-import org.edumips64.core.*;
-import org.edumips64.core.fpu.*;
-//per diagnostica
+import org.edumips64.core.Converter;
+import org.edumips64.core.FCSRRegister;
+import org.edumips64.core.IrregularStringOfBitsException;
+import org.edumips64.core.IrregularWriteOperationException;
+import org.edumips64.core.MemoryElementNotFoundException;
+import org.edumips64.core.fpu.FPDivideByZeroException;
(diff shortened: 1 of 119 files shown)


src/main/java/org/edumips64/core/is/FPArithmeticInstructions.java
```diff
@@ -24,10 +24,19 @@
...
+import org.edumips64.core.IrregularWriteOperationException;
+import org.edumips64.core.MemoryElementNotFoundException;
+import org.edumips64.core.fpu.FPDivideByZeroException;
+import org.edumips64.core.fpu.FPInstructionUtils;
+import org.edumips64.core.fpu.FPInvalidOperationException;
+import org.edumips64.core.fpu.FPOverflowException;
+import org.edumips64.core.fpu.FPUnderflowException;
+import org.edumips64.core.fpu.RegisterFP;
 
+//per diagnostica
```
lupino3, line 39: You can remove this comment, it looks obsolete.

lupino3:
Hi, thanks a lot! The change looks good. There is an obsolete comment which you moved around: can you please delete it? Other than that, the change looks great and I'll merge it once you remove the comment.

If you prefer not to remove the comment, just let me know and I'll merge the change as it is.

Thanks for your contribution!

src/main/java/org/edumips64/core/is/FPArithmeticInstructions.java
pimts: Sure! No problem, shall I remove all occurrences of that comment across all classes?

lupino3:
Thanks a lot @pimts for the change and for the clean-ups! Merging the change.

There is one check that's failing, but it's not failing because of your change.

PR #448, opened 2020-08-09
Remove ParsedArgs from Main and replace with new EduMipsArgs.

iwodder:
@lupino3 ready for you to review, let me know if you want any changes. I think this should meet the issue reqs.


lupino3:
Hi @iwodder,
I tried running the JAR locally and realized it seems to crash when started with `-d`:
image

(I got the JAR from the CI build)

Did you try running it locally with some combination of arguments?

Unfortunately I have a very busy week at work, I hope I'll be able to review the change in the weekend.

Thanks,
Andrea

iwodder:
@lupino3 oops, I did test it with a variety of args. I must have missed that option on Main.java, all args worked on MainCLI.java. I found the issue and have corrected locally. Before pushing up, I noticed there is an issue. That NPE comes because the configStore is not initialized before trying to invoke resetConfiguration. Is the way that MainCLI initializes the configStore the same way to do it in Main?

lupino3:
I think we should try to initialize those core classes as soon as possible in the constructor of `Main`.

I have some ideas (of varying complexity) on how to improve the situation here, the existing code is not great and prone to errors, but in general the easiest way, I think, is to initialize in the constructor of `Main` ASAP.

*ideally* we should use dependency injection. There are 3 flavors of EduMIPS64:

- GUI
- CLI
- Web

They all need the same types of components but might need different implementations (specifically the I/O and the preferences, if I remember correctly, are different). We could abstract these away and use DI to inject them when we construct the "main" object (which really is a `Simulator`, an abstraction we don't have).

For the scope of this change, I'd suggest to just try to move the initialization in the correct place.

Thank you!

src/main/java/org/edumips64/utils/args/EduMipsArgs.java
```diff
@@ -0,0 +1,48 @@
...
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+
+@Command(mixinStandardHelpOptions = true, resourceBundle = "Messages", versionProvider = EduMipsVersion.class)
+public class EduMipsArgs implements Runnable {
```
lupino3, line 13: I have a small naming comment. As far as I remember, we always tried to avoid having `EduMIPS` or similar in the class names, because it's already in the package name. Therefore, could you please rename this class to `Args` and `EduMipsVersion` to `Version`?

Another minor suggestion: I'd suggest to use the `org.edumips64.utils.cli` package rather than `org.edumips64.utils.args`, to convey a bit more the idea that those classes are related to the command-line invocation.

src/main/java/org/edumips64/utils/args/EduMipsVersion.java
```diff
@@ -0,0 +1,16 @@
+package org.edumips64.utils.args;
+
+import org.edumips64.utils.MetaInfo;
+import picocli.CommandLine;
+
+public class EduMipsVersion implements CommandLine.IVersionProvider {
+
+    @Override
+    public String[] getVersion() {
+        return new String[]{"EduMIPS64 version " + MetaInfo.VERSION +
```
lupino3, line 10: Would it be possible to move the code that generates this string to a static method?

The reason is that to get the text the new code creates a new `EduMipsVersion` instance just to call `getVersion()` and get this string, while the whole behavior doesn't really need the allocation of a new version, it could be something like  `EduMipsVersion.getVersionString();`.

iwodder: I will try that in the updates I am working on for this branch. Picocli requires using this style to allow dynamically generating the version, but I don't think it would break to add a static method. Will let you know if it doesn't work, otherwise will update.

src/main/java/org/edumips64/utils/args/EduMipsArgs.java
iwodder: Understood, will make those changes for this.

src/main/java/org/edumips64/utils/args/EduMipsVersion.java
lupino3: Sorry, I was not sufficiently clear. What I meant is that we can move the logic itself in a static method (or a `public static final string` really), and use it in `getVersion()`.

iwodder:
@lupino3 updated with changes, ready for another look.

src/main/java/org/edumips64/utils/args/EduMipsVersion.java
lupino3: Looks good, thanks!

src/main/java/org/edumips64/utils/args/EduMipsArgs.java
lupino3: Looks good, thanks!

lupino3:
Thanks a lot for the contribution. Looks great, merged!

PR #454, opened 2020-08-27
Make cli more useable

iwodder:
@lupino3 ready for a first look here. I took care of most things from the open issue.

- Added file command to open file from CLI
- Added Dinero tracefile output (made an attempt based on what I saw in the GUI source)
- Added config command to print current configuration
- Added i18n

If this gets signed off and merged, then I can go back and complete the remaining items for documentation and non-interactive mode. I think the editing configuration would be quite large but doable, if you would like that done I can do that too.

### DIFF
--- a/src/main/java/org/edumips64/MainCLI.java
+++ b/src/main/java/org/edumips64/MainCLI.java
@@ -38,10 +38,15 @@ import org.edumips64.utils.CurrentLocale;
 import org.edumips64.utils.JavaPrefsConfigStore;
 import org.edumips64.utils.MetaInfo;
 import org.edumips64.utils.cli.Args;
+import org.edumips64.utils.cli.Cli;
 import org.edumips64.utils.io.LocalFileUtils;
 import picocli.CommandLine;
 
 import java.io.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.ResourceBundle;
 
 /** Interactive shell for EduMIPS64
  * @author Andrea Spadaccini
@@ -62,146 +67,31 @@ public class MainCLI {
         cfg.resetConfiguration();
       }
 
-      // Initialize the CPU and all its dependencies.
-      Memory memory = new Memory();
-      CPU c = new CPU(memory, cfg, new BUBBLE());
-      LocalFileUtils localFileUtils = new LocalFileUtils();
-      SymbolTable symTab = new SymbolTable(memory);
-      IOManager iom = new IOManager(localFileUtils, memory);
-      Dinero dinero = new Dinero();
-      InstructionBuilder instructionBuilder = new InstructionBuilder(memory, iom, c, dinero, cfg);
-      Parser p = new Parser(localFileUtils, symTab, memory, instructionBuilder);
-      c.setStatus(CPU.CPUStatus.READY);
+      Cli cli = new Cli(cfg);
+      commandLine = new CommandLine(cli);
 
+      System.out.println(CurrentLocale.getString("CLI.WELCOME"));
       // Initialization done. Print a welcome message and open the file if needed.
-      System.out.println("Welcome to EduMIPS64 CLI shell!");
       if (cliArgs.getFileName() != null) {
-        String absoluteFilename = new File(cliArgs.getFileName()).getAbsolutePath();
-        try {
-          p.parse(absoluteFilename);
-        } catch (ParserMultiException e) {
-          if (e.hasErrors()) {
-            throw e;
-          }
-        }
-        dinero.setDataOffset(memory.getInstructionsNumber()*4);
-        c.setStatus(CPU.CPUStatus.RUNNING);
-        System.out.println("(Loaded file " + absoluteFilename + ")");
+          commandLine.execute("file", cliArgs.getFileName());
       }
 
-      // Start the (very primitive) Read/Eval/Print Loop.
-      BufferedReader keyboard = new BufferedReader(new InputStreamReader(System.in));
-      System.out.print("> ");
-      while (true) {
+      runReplLoop(commandLine);
 
-        String read = keyboard.readLine();
-
-        if (read == null || read.equals("exit")) {
-          break;
-        }
-
-        String[] tokens = read.split(" ");
-
-        if (tokens[0].compareToIgnoreCase("help") == 0) {
-          String help = "EduMIPS64 CLI SHELL - Available commands:\n";
-          help += "-----------------------------------------\n";
-          help += "help\t\t\tshow this help message\n";
-          help += "step\t\t\tmake the CPU state machine advance of one step\n";
-          help += "step n\t\t\tmake the CPU state machine advance of n steps\n";
-          help += "run\t\t\texecute the program till its end\n";
-          help += "show registers\t\tshow the content of registries\n";
-          help += "show memory\t\tshow the content of memory\n";
-          help += "show symbols\t\tshow the content of the symbol table\n";
-          help += "show pipeline\t\tshow the content of the pipeline\n";
-          help += "exit\t\t\texit EduMIPS64 CLI shell\n";
-          System.out.println(help);
-        } else if (tokens[0].compareToIgnoreCase("show") == 0) {
-          if (tokens.length == 1) {
-            System.out.println("The show command requires at least one parameter.");
-          } else {
-            if (tokens[1].compareToIgnoreCase("registers") == 0) {
-              System.out.println(c.gprString());
-            } else if (tokens[1].compareToIgnoreCase("register") == 0) {
-              if (tokens.length < 3) {
-                System.out.println("The show register command requires at least one parameter.");
-              } else {
-                System.out.println(c.getRegister(Integer.parseInt(tokens[2])));
-              }
-            } else if (tokens[1].compareToIgnoreCase("memory") == 0) {
-              System.out.println(memory);
-            } else if (tokens[1].compareToIgnoreCase("symbols") == 0) {
-              System.out.println(symTab);
-            } else if (tokens[1].compareToIgnoreCase("pipeline") == 0) {
-              System.out.println(c.pipeLineString());
-            } else {
-              System.out.println("The show command requires at least one parameter.");
-            }
-          }
-        } else if (tokens[0].compareTo("run") == 0) {
-            int steps = 0;
-            long startTimeMs = System.currentTimeMillis();
-            try {
-              while(true) {
-                steps++;
-                c.step();
-
-              }
-            } catch (HaltException e) {
-              long endTimeMs = System.currentTimeMillis();
-              long totalTimeMs = endTimeMs - startTimeMs;
-              System.out.println("Execution ended. " + steps + " steps were executed in " + totalTimeMs + "ms");
-            }
-        } else if (tokens[0].compareTo("step") == 0) {
-          try {
-            int num = 1;
-
-            if (tokens.length > 1) {
-              num = Integer.parseInt(tokens[1]);
-            }
-
-            try {
-              if (num > 0) {
-                System.out.println("I execute " + num + " steps of simulation.");
-              }
-
-              for (int i = 0; i < num; ++i) {
-                c.step();
-                System.out.println(c.pipeLineString());
-              }
-            } catch (Exception e) {
-              System.err.println(getErrorReportingMessage());
-              e.printStackTrace();
-            }
-          } catch (NumberFormatException e) {
-            System.out.println("The second parameter of the step command must be an integer.");
-          }
-        } else {
-          System.out.println("Unknown command.\nType 'help' to get the list of available commands.");
-        }
-
-        System.out.print("> ");
-      }
-
-      System.out.println("Bye!");
     } catch (Exception e) {
-      System.err.println(getErrorReportingMessage());
-      e.printStackTrace();
+      Cli.printErrorMessage(e);
       System.exit(1);
     }
   }
-  
-  /**
-   * Compiles and returns a generic error message with version 
-   * and system diagnostic information to be used to report the error.
-   * @return A string containing the error message.
-   */
-  private static String getErrorReportingMessage() {
-    String msg = "EduMIPS64 fatal error!\n" +
-        "Please report the following stacktrace and system information,\n" +
-        "along with the content of the assembly file you were executing\n" +
-        "to the EduMIPS64 GitHub account: https://github.com/lupino3/edumips64/issues/new\n";
-    msg += String.format("Version: %s, %s, %s\n", MetaInfo.VERSION, MetaInfo.BUILD_DATE, MetaInfo.FULL_BUILDSTRING);
-    msg += String.format("JRE version: %s\nOS: %s\n\n", System.getProperty("java.version"), System.getProperty("os.name"));
-    return msg;
-  }  
+
+    private static void runReplLoop(CommandLine commandLine) throws IOException {
+        BufferedReader keyboard = new BufferedReader(new InputStreamReader(System.in));
+        System.out.print("> ");
+        while (true) {
+          String read = keyboard.readLine();
+          String[] tokens = read.split(" ");
+          commandLine.execute(tokens);
+          System.out.print("> ");
+        }
+    }
 }
--- a/src/main/java/org/edumips64/utils/cli/Cli.java
+++ b/src/main/java/org/edumips64/utils/cli/Cli.java
@@ -33,7 +33,6 @@ public class Cli implements Runnable {
 
     private final Memory memory;
     private final CPU c;
-    private final LocalFileUtils localFileUtils;
     private final SymbolTable symTab;
     private final IOManager iom;
     private final Dinero dinero;
@@ -45,12 +44,11 @@ public class Cli implements Runnable {
         this.configStore = cfg;
         memory = new Memory();
         c = new CPU(memory, cfg, new BUBBLE());
-        localFileUtils = new LocalFileUtils();
         symTab = new SymbolTable(memory);
-        iom = new IOManager(localFileUtils, memory);
+        iom = new IOManager(new LocalFileUtils(), memory);
         dinero = new Dinero();
         instructionBuilder = new InstructionBuilder(memory, iom, c, dinero, cfg);
-        p = new Parser(localFileUtils, symTab, memory, instructionBuilder);
+        p = new Parser(new LocalFileUtils(), symTab, memory, instructionBuilder);
         c.setStatus(CPU.CPUStatus.READY);
     }
 
@@ -76,10 +74,6 @@ public class Cli implements Runnable {
 
     public CPU getCPU() {
         return c;
-    }
-
-    public LocalFileUtils getLocalFileUtils() {
-        return localFileUtils;
     }
 
     public SymbolTable getSymTab() {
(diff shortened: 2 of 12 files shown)


lupino3:
Hi @iwodder, thanks a lot for the PR! The code overall looks good, I will try it out locally before merging it.

I have some minor suggestions for the PR, I'll send you the comments as soon as possible, but in the meantime I wanted to thank you and let you know that I like the overall approach and the benefits that this PR brings!

src/main/java/org/edumips64/utils/cli/StepCommand.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void run() {
+        try {
+            if (steps > 0) {
+                printStepMsg();
+            }
+
+            for (int i = 0; i < steps; ++i) {
+                cli.getCPU().step();
```
lupino3, line 29: This will throw an exception if the CPU is not in the right step (try loading the JAR and trying to `step` without opening a program).

You should check for the CPU status, similarly to this: https://github.com/EduMIPS64/edumips64/blob/master/src/main/java/org/edumips64/client/Simulator.java#L85

I recognize that this is sub-optimal and the interface is not great... could be easier! The CPU could provide a method to let the caller know that it can run code. But to be fair the CPU class was never designed with multiple possible simulator implementations!

src/main/java/org/edumips64/MainCLI.java
```diff
@@ -62,146 +67,31 @@ public static void main(String args[]) {
...
-    msg += String.format("Version: %s, %s, %s\n", MetaInfo.VERSION, MetaInfo.BUILD_DATE, MetaInfo.FULL_BUILDSTRING);
-    msg += String.format("JRE version: %s\nOS: %s\n\n", System.getProperty("java.version"), System.getProperty("os.name"));
-    return msg;
-  }  
+
+    private static void runReplLoop(CommandLine commandLine) throws IOException {
+        BufferedReader keyboard = new BufferedReader(new InputStreamReader(System.in));
+        System.out.print("> ");
+        while (true) {
+          String read = keyboard.readLine();
```
lupino3, line 91: It might be worth to not invoke `execute` if the input is empty or contains only whitespace, as that will return to the user a non-informative user message (`Unmatched argument at index 0: ''`)

src/main/java/org/edumips64/utils/cli/ShowCommand.java
```diff
@@ -0,0 +1,46 @@
+package org.edumips64.utils.cli;
+
+import picocli.CommandLine.Command;
+import picocli.CommandLine.ParentCommand;
+import picocli.CommandLine.Parameters;
+import picocli.CommandLine.Option;
+
+@Command(name = "show", resourceBundle = "CliMessages")
+public class ShowCommand implements Runnable{
```
lupino3, line 9: Would it be possible to print the help if the user invokes `show` with no arguments? Currently it gives no feedback, which could be confusing.

src/main/java/org/edumips64/utils/cli/Cli.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Compiles and returns a generic error message with version
+     * and system diagnostic information to be used to report the error.
+     * @return A string containing the error message.
+     */
+    public static String getErrorReportingMessage() {
+        String msg = "EduMIPS64 fatal error!\n" +
+                "Please report the following stacktrace and system information,\n" +
+                "along with the content of the assembly file you were executing\n" +
+                "to the EduMIPS64 GitHub account: https://github.com/lupino3/edumips64/issues/new\n";
```
lupino3, line 123: Note that the URL changed now, it's `https://github.com/EduMIPS64/edumips64/issues/new` (I moved the project from my personal account to an organization recently)

src/main/java/org/edumips64/utils/cli/Cli.java
```diff
@@ -0,0 +1,128 @@
...
+    private SymbolTable symTab;
+    private IOManager iom;
+    private Dinero dinero;
+    private InstructionBuilder instructionBuilder;
+    private Parser p;
+    private final ConfigStore configStore;
+
+    public Cli(ConfigStore cfg) {
+        this.configStore = cfg;
+        memory = new Memory();
```
lupino3, line 39: Could you just invoke `reset()` to do the initialization here, right after initializing `this.configStore`?

src/main/java/org/edumips64/utils/cli/ConfigCommand.java
```diff
@@ -0,0 +1,39 @@
...
+    private static ConfigStore configStore;
+
+    @Override
+    public void run() {
+        configStore = cli.getConfigStore();
+        System.out.println(CurrentLocale.getString("CLI.CONFIG.HEADER"));
+        Stream.of(ConfigKey.values()).forEach(ConfigCommand::printValue);
+    }
+
+    private static void printValue(ConfigKey k) {
```
lupino3, line 26: Nice way to print the value not knowing the type! :) I think `ConfigStore` could have a way to make this easier (for example a `printValue(ConfigKey key)` method, but since this is the only use case I can foresee for such a method I think that the workaround you have here is absolutely fine!

src/main/java/org/edumips64/utils/cli/DineroCommand.java
```diff
@@ -0,0 +1,54 @@
...
+        } catch (FileNotFoundException fnfe) {
+            System.out.println(CurrentLocale.getString("CLI.FILE.NOT.FOUND"));
+        } catch (IOException ioe) {
+            System.out.println(CurrentLocale.getString("CLI.IO.ERROR"));
+        } catch (Exception e) {
+            Cli.printErrorMessage(e);
+        }
+    }
+
+    private String appendDineroSuffix(String s) {
```
lupino3, line 47: Why is this needed? I think it might be surprising for the user if they ask the simulator to write the tracefile to `foo.txt` and we write to `foo.txt.xdin`.

src/main/java/org/edumips64/utils/cli/FileCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import org.edumips64.utils.io.ReadException;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.ParentCommand;
+import picocli.CommandLine.Parameters;
+import picocli.CommandLine.Option;
+
+import java.io.File;
+
+
+@Command(name = "file", resourceBundle = "CliMessages")
```
lupino3, line 15: I'd suggest to rename `FileCommand` to `LoadCommand`, since "load" to me is the action that the command performs. (or "open", maybe?)

src/main/java/org/edumips64/utils/cli/FileCommand.java
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            if (canLoadFile()) {
+                cli.getParser().parse(absoluteFilename);
+                cli.getDinero().setDataOffset(cli.getMemory().getInstructionsNumber() * 4);
+                cli.getCPU().setStatus(CPU.CPUStatus.RUNNING);
+                printLoadedFileMsg(absoluteFilename);
+            } else {
+                printCannotLoad();
+            }
+        } catch (ParserMultiException | ReadException e) {
```
lupino3, line 39: The `ParserMultiException` contains all the parsing errors encountered by the simulator, and they are necessary for the user to understand how to fix their code.

Would it be possible to output the parsing errors?

src/main/java/org/edumips64/utils/cli/ShowCommand.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @ParentCommand
+    private Cli cli;
+
+    @Option(names={"-h","--help"}, usageHelp = true, descriptionKey= ".help.usage.description")
+    private boolean help;
+
+    @Command(name="registers")
+    private void showRegisters() {
```
lupino3, line 18: Would it be possible to add commands to see the FP and special registers (LO/HI/FCSR)?

src/test/java/org/edumips64/MainCLITest.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    private OutputStream getSystemErr() {
+        ByteArrayOutputStream boas = new ByteArrayOutputStream();
+        System.setErr(new PrintStream(boas));
+        return boas;
+    }
+
+    private String enMainUsageMsg() {
+        return "Usage:  [COMMAND]\n" +
```
lupino3, line 169: I'd suggest to not encode the full output of the help here, because as the functionality evolves (adding new commands, changing parameters, etc.) we as authors will need to continuously update these tests to make sure they are up-to-date.

src/main/resources/CliMessages.properties
```diff
@@ -0,0 +1,37 @@
...
+
+# Dinero command messages
+.dinero.usage.description=Write dinero tracefile to a file
+dinero.param=The file to write to
+
+# Run command messages
+.run.usage.description=Execute the program without intervention
+
+# Show command messages
+.show.usage.description=Show various content of the CPU state machine
```
lupino3, line 20: Why are you using both `CurrentLocale.java` and properties files to encode localized text?

lupino3 (changes requested):
Thank you very much for the changes. I tested the simulator locally and it is *so much better* than it used to be!

I have a few change requests for you, happy to discuss any of them or to implement some of them in a subsequent PR.

Thanks again!

iwodder:
@lupino3 thanks for the feedback, I am happy to help. I will get this update in the next couple days.

lupino3:
Thank you very much!

src/main/java/org/edumips64/utils/cli/Cli.java
iwodder: Got it, updated.

src/main/java/org/edumips64/utils/cli/Cli.java
iwodder: Yeah, that's a good call. Done.

src/main/resources/CliMessages.properties
iwodder: I was trying to stay consistent and add messages which are output directly from the source in the same manner currently in use. The resource bundles are only used for the PicoCli lib right now. However, I was (forgot to... oops), going to open an issue for converting the entire CurrentLocale to use a resource bundle like the PicoCli lib uses. That way it will be consistent. Thoughts?

src/main/java/org/edumips64/utils/cli/ConfigCommand.java
iwodder: Thanks, I looked for a convenience method there but didn't see it. That would probably be a good add. Of course, I could be looking in the wrong spot

src/main/java/org/edumips64/utils/cli/FileCommand.java
iwodder: I think load sounds good. Changed to that.

src/main/java/org/edumips64/utils/cli/FileCommand.java
iwodder: Definitely, looking inside ParserMultiException, it appears there is a good toString(). Is that preferred for outputting the parser errors?

src/test/java/org/edumips64/MainCLITest.java
iwodder: Good call, updated to just check sysout is not empty.

src/main/java/org/edumips64/utils/cli/DineroCommand.java
iwodder: I did this because in the GUI when the user wants to write out a tracefile it is presented with that suffix in the file dialog. Do you want me to change or leave as is? I realize in the GUI the user can change the extension in the file dialog but here I would imagine a CLI user is more saavy and could just rename the file if they wanted a different name.

src/main/java/org/edumips64/utils/cli/ShowCommand.java
iwodder: Yup, changed it!

src/main/java/org/edumips64/MainCLI.java
iwodder: I agree. I would think that printing the help message would be ideal for entering no input, thoughts?

lupino3: Yeah, makes sense! Thanks!

src/main/resources/CliMessages.properties
lupino3: Makes sense to me! I would prefer to have a single way to store localized strings, and I am perfectly ok to have an intermediate state where the messages are split and an issue to converge to a single mechanism, be it the Java file or the resource bundles. Of the two, of course I prefer the resource bundle, it's more elegant :) thank you

src/test/java/org/edumips64/MainCLITest.java
lupino3: Could you please push your changes so I can see the updated code?

src/main/java/org/edumips64/utils/cli/FileCommand.java
lupino3: I think that should be ok, yes! We might need to tweak it a bit (not sure if that correctly distinguishes warnings vs. errors), I'd need to test it to be sure :) But as a first step, that's definitely good enough!

src/main/java/org/edumips64/utils/cli/DineroCommand.java
lupino3: I understand what you mean. I do not fully agree (I'd prefer users to fully control the output file name), but I accept your choice!

Can you please echo back to the user the actual filename the code used, so that they will know whether we added the extension or not?

src/test/java/org/edumips64/MainCLITest.java
iwodder: Pushed up most recent changes just now. Still working on some of the other comments you had.

src/main/java/org/edumips64/utils/cli/StepCommand.java
iwodder: Ahh.. I see. I will modify it to check for the status!

src/main/java/org/edumips64/utils/cli/ShowCommand.java
iwodder: Added in the most recent commit, incoming here shortly.

src/main/java/org/edumips64/utils/cli/DineroCommand.java
iwodder: Added the echo for the file name. Removed the extension, it was easier to echo without adding.

iwodder:
@lupino3 Most update to branch pushed, ready for another check!

src/main/java/org/edumips64/MainCLI.java
```diff
@@ -62,146 +67,46 @@ public static void main(String args[]) {
...
+    System.out.print("> ");
+  }
 
-      System.out.println("Bye!");
-    } catch (Exception e) {
-      System.err.println(getErrorReportingMessage());
-      e.printStackTrace();
-      System.exit(1);
+  private static void handleTokens(CommandLine commandLine, String[] tokens) {
+    if (tokens.length > 0) {
```
lupino3, line 102: I tried the last JAR and it still seems to show the error before showing the usage:
image

Maybe `tokens` here is not empty but has empty/null entries?

This is definitely a minor thing though. I wouldn't block merging on this one.

src/main/java/org/edumips64/MainCLI.java
```diff
@@ -62,146 +67,46 @@ public static void main(String args[]) {
...
-          p.parse(absoluteFilename);
-        } catch (ParserMultiException e) {
-          if (e.hasErrors()) {
-            throw e;
-          }
-        }
-        dinero.setDataOffset(memory.getInstructionsNumber()*4);
-        c.setStatus(CPU.CPUStatus.RUNNING);
-        System.out.println("(Loaded file " + absoluteFilename + ")");
+          commandLine.execute("file", cliArgs.getFileName());
```
lupino3, line 76: The command is `load` now, right?

```suggestion
          commandLine.execute("load", cliArgs.getFileName());
```

src/main/java/org/edumips64/utils/cli/RunCommand.java
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            cli.printNotRunnable();
+        }
+    }
+
+    private void execute() {
+        int steps = 0;
+        long startTimeMs = System.currentTimeMillis();
+        try {
+            while(true) {
```
lupino3, line 27: Programs can run for thousands of steps or more.

1. can you please add some sort of visual feedback to the user? For example, a dot every 100 steps or so?
2. how can the user stop the execution? Do they use SIGINT or other signals? If so, what happens? Do we want the emulator to retain control of the execution (to allow the user to inspect what happened) or do we want the program to terminate?

I think point 2 can be considered out of scope for this PR, but I'd prefer you to address point 1 (visual feedback) if possible.

src/main/java/org/edumips64/utils/cli/RunCommand.java
```diff
@@ -0,0 +1,44 @@
...
+                steps++;
+                cli.getCPU().step();
+            }
+        } catch (HaltException e) {
+            long endTimeMs = System.currentTimeMillis();
+            long totalTimeMs = endTimeMs - startTimeMs;
+            printEndOfExecMsg(steps, totalTimeMs);
```
lupino3, line 34: Nice to give some feedback about the execution time! :)

src/main/java/org/edumips64/utils/cli/ShowCommand.java
lupino3: I just tried it out, and it work.

I am ok with submitting as it is, but I have a few suggestion for a future iteration:
- the `show register` command could be the entry point to see all registers, including single FP registers and special registers. It'd be nice to support all the names supported by the parser (e.g. `show register r1`, `show register f12`, `show register lo`) including aliases (`show register $a0`, etc...)

Again, it's ok as it is now, just writing things down for future improvements!

lupino3 (commented):
Thanks, I think it's in good shape!

I am happy to merge once you address the 2 main points raised in the last iteration:
- use `load` instead of `file` in `MainCLI`
- give some visual feedback to users as they run long commands.

Thanks a lot for this large improvement!

src/main/java/org/edumips64/utils/cli/RunCommand.java
iwodder: @lupino3 will resolve that today or tomorrow and push up the change.

src/main/java/org/edumips64/utils/cli/RunCommand.java
iwodder: Thanks!

iwodder:
@lupino3 ready for another check. I changed up run to print out a message indicating execution started. Then it will print "Running" followed with a dot every 100 steps.

lupino3 (approved):

lupino3:
LGTM, merged! Thanks @iwodder !!

lupino3:
(the main issue I see is the splash screen feature.... we might need to remove it or work around it, as it might crash when used from a terminal with no X)

iwodder:
I think that is a good idea and it would consolidate the packaging. I don't think it would be too hard to manage the splash screen. We would need to remove it from the Manifest file and programmatically display it instead.

lupino3:
Created https://github.com/EduMIPS64/edumips64/issues/457 to track the consolidation, thanks!

Let me know if you would like to work on it! I'd really appreciate your help on this :)

iwodder:
@lupino3 I saw that new issue for consolidation. I can work on that this weekend. That will be a great improvement.

Issue #302, opened 2020-04-05
Create Windows installer


lupino3:
There's a StackOverflow question on how to start EduMIPS64 in Windows: https://stackoverflow.com/questions/60990995/how-can-i-open-edumips-jar/61039766#61039766

For people not familiar with JAR files, this may not be super-intuitive.

We should explore the possibility of creating an installer for Windows, ideally bundling a JDK with it and building the `exe` (or `msi`) from CI.

lupino3:
From a cursory look, launch4j seems promising: http://launch4j.sourceforge.net/

lupino3:
There are some examples here on how to use a Gradle task to generate an executable with bundled JDK: https://github.com/TheBoegl/gradle-launch4j/issues/86

lupino3:
I can't really find motivation to finish this task, I'll probably work on it if enough people request an installer / Windows binary.

rocas777:
Hey there, I was able to make an executable with launch4J. Can make it work with the exe file and a folder with the jre. Since the exe refers to the jre with a relative path,you can make a release with only those two files. Then pack it with NSIS and give it away to Windows Users.

lupino3:
Thanks @rocas777! That sounds like what we need!

Just one small thing: the point of the Windows installer would be to offer a "single click" installation experience, so I'd like to bundle the JRE within the installer.

Could you please make that change before sending the PR with the launch4j config?

Thanks for your help!

lupino3:
Another alternative might be to use the tooling available in JDK 14+: https://openjdk.java.net/jeps/343

PR #464, opened 2020-10-01
Windows Installer automatic creator

rocas777:
This scrips adds the possibility to create an msi installer.(only works on Windows)
It depends on Wix and the jdk version must be 14+;
To create the installer run the command
" gradlew.bat exeCreator "

The script creates the jar, documentation and msi.


rocas777:
closes #302

build.gradle.kts
```diff
@@ -1,214 +1,259 @@
...
+            throw GradleException("Please Install Wix")
+        }
+        else{
+            println("Wix Installed!")
+        }
+
+	//create script file that creates the executable
+        var script = File("./scripts/exeCreator.bat")
+        script.createNewFile()
+        var version_arr = version.split('.').toMutableList();
```
lupino3, line 242: Is this code necessary to truncate the version number? If so, could you please add a comment and emit a warning to the user that the version number is truncated?

build.gradle.kts
```diff
@@ -1,214 +1,259 @@
...
+        if(version_arr.size>3 && version_arr[3].toInt()>0) {
+            version_arr[2] = (version_arr[2].toInt()+ 1).toString()
+        }
+        var jpackage_version = version_arr[0]+'.'+version_arr[1]+'.'+version_arr[2]
+        println("Running JPackage!")
+        script.writeText("jpackage.exe --main-jar edumips64-"+version+".jar --input ./build/libs/ --app-version "+jpackage_version+" --name EduMIPS64 --description \"Educational MIPS64 CPU Simulator\" --vendor \"EduMIPS64 Development Team\" --copyright \"Copyright "+LocalDateTime.now().year+", EduMIPS64 development Team\" --license-file ./LICENSE --win-shortcut --win-dir-chooser --win-menu --type msi --icon ./src/main/resources/images/ico.ico --win-per-user-install")
```
lupino3, line 248: Instead of creating an intermediate `.bat` file, could we directly invoke `jpackage.exe` from the Gradle task itself?

build.gradle.kts
```diff
@@ -1,214 +1,259 @@
...
+    description = "Creates all artifacts for a given EduMIPS64 release"
+    dependsOn("allDocs")
+    dependsOn("jar")
+
+    doFirst {
+        println("Creating artifacts for version $version")
+    }
+}
+
+tasks.create<Exec>("exeCreator"){
```
lupino3, line 208:
```suggestion
tasks.create<Exec>("createMsi"){

```

Since this creates an MSI installer, I suggest to call it `createMsi` or something similar.

lupino3 (changes requested):
Thanks a lot for the change! I have a couple small change requests, but the approach overall looks good to me.

build.gradle.kts
rocas777: I followed the pattern on the jpackage command you sent me.
e.g. If version is 1.2.7.1, jpackage version is 1.2.8, jpackage doesn't allow the first version model.
The separation in a list is mandatory to allow the last number to have more than 1 digit

build.gradle.kts
rocas777: For some reason the jpackage.exe command is not recognized if called directly from the gradle tasks. Probably it doesn't import the env variables. 
I will try other approach

rocas777: I can't seem to get it done. Already tried to specify the full path and still does not work. 
I can add  
```java
    catch (ex:Exception){
        script.delete();
    }
    doLast{
        script.delete();
    }
```
to erase the script file once it is no longer needed. (a finally block should work but doesn't)
This way the file is deleted either the build is successful or not.
Is the best i can do.

build.gradle.kts
lupino3: 1.2.7.1 is a very special case, we can assume I'll never use the 4th number and simplify the code so that it just uses the version number.

I'll bump up the version to 1.2.8 as soon as possible.

It's OK if you just add a TODO comment and I can simplify it once I bump up the version to 1.2.8.

build.gradle.kts
lupino3: That's ok, no problem at all. You can leave it as it is and we may investigate it later, for now it works and it's a great new feature!

lupino3 (approved):
Looks good, thank you!

build.gradle.kts
lupino3: I filed https://github.com/EduMIPS64/edumips64/issues/465 to keep track of this clean-up, looks good as it is!

Issue #489, opened 2020-10-23
Character encoding issues in manual


pviotti:
Some characters in the manual are badly rendered. 

![image](https://user-images.githubusercontent.com/1350095/97033379-2ad8f180-155b-11eb-9738-214fbe81da2b.png)
![image](https://user-images.githubusercontent.com/1350095/97033419-3a583a80-155b-11eb-9bba-3c9ecd2e6a5d.png)
etc.

v1.2.8 on Windows and Linux.

lupino3:
Thanks @pviotti! Curiously enough, it all seems ok on readthedocs: https://edumips64.readthedocs.io/en/latest/

lupino3:
![Screenshot 2020-10-24 at 19 00 41](https://user-images.githubusercontent.com/200139/97088800-44e20500-162b-11eb-9269-7898f293704a.png)
![Screenshot 2020-10-24 at 19 00 26](https://user-images.githubusercontent.com/200139/97088802-4ad7e600-162b-11eb-8881-c8996f7b8dce.png)

Also works on Mac OS X

lupino3:
![image](https://user-images.githubusercontent.com/200139/97088864-b621b800-162b-11eb-9ca5-3ad0b47395b5.png)
The issue is visible even for the JAR built by CI (just excluding a possible misconfiguration of my local desktop where I built the JAR)

lupino3:
Broken even in 1.2.7:
![image](https://user-images.githubusercontent.com/200139/97088985-7d361300-162c-11eb-81c6-96613da117ff.png)

lupino3:
1.2.6 looks better, so the breakage is likely to have happened between 1.2.6 and 1.2.7:
![image](https://user-images.githubusercontent.com/200139/97089001-a6ef3a00-162c-11eb-8b88-4371fcff6cb8.png)

lupino3:
BTW 1.2.8 does not look broken to me on Ubuntu 20.04
![Screenshot from 2020-10-24 21-24-40](https://user-images.githubusercontent.com/200139/97092879-6a790980-163f-11eb-8ae9-059056fea32d.png)
@pviotti what version of Linux are you using?

pviotti:
[...]

PR #491, opened 2020-10-25
Replace javahelp with default browser

pviotti:
JavaHelp [hasn't been actively maintained since 2010](https://github.com/javaee/javahelp/commits/master), and it's causing some issues (#489 ).  
This PR attempts to replace the JavaHelp manual window with a launcher for the default application that reads HTML files.  
HTML files of the manual in the jar are lazily decompressed to a temporary folder the first time the manual is opened (this is needed because browsing a set of HTML files inside a jar file [is not feasible](https://stackoverflow.com/a/10187565)).

To do / fix:
 - [x] opening the manual when executing not from a jar throws an exception because cross-platform/jar Path API in Java is gnarly (see comment in the code)... ([ref1](https://stackoverflow.com/questions/43530119/error-when-using-filesystems-newfilesystemuri-uri-mapstring-env) [ref2](https://stackoverflow.com/questions/27078996/path-component-should-be) [ref3](https://stackoverflow.com/questions/11682907/not-too-sure-how-a-uri-works-regarding-absolute-paths-to-files) - not sure I can fix this)
 - [x] remove all remaining JavaHelp references and build files
 - [x] support Italian manual
 - [ ] test on Windows (can't test on MacOS)


lupino3:
I just tried it out on Mac OS X and it works, thanks! The only thing that doesn't seem to work is opening the help in Italian: even if I change the language it still opens the English one.

I was wondering if it would be worth to try out if it's possible to replace JavaHelp with a Swing widget that can display the help: looks like `JTextPane` can do it. Doing it would make EduMIPS64 able to display its own help.

What do you think? I definitely want to get rid of JavaHelp. But I also think it'd be nice to not lose the ability to display the help without external dependencies.

lupino3:
If we end up adding a way to display the help in-app, it might be worth going with the [JavaFX WebView](https://docs.oracle.com/javafx/2/webview/jfxpub-webview.htm) rather than hand-crafting things with JTextPane or similar though, because the help is on multiple HTML files. Another alternative might be to use the simpler JTextPane approach but output the help as a single HTML file, so we don't have to deal with following hyperlinks which I assume will be challenging using JTextPane.

pviotti:
> opening the help in Italian

it should be fixed now (haven't tested it).

> but I also think it'd be nice to not lose the ability to display the help without external dependencies.

yeah that's true. On the other hand:
 - probably with any implementation based on JavaFX or JTextPane it would be non-trivial to have basic functionalities such as navigation and text search, which are a given in every browser, and familiar to the user too (btw I just realized current in-app manual doesn't have text search either...)
 - it would require a few (~100?) more lines of code to maintain and test
 - browsers are installed on every machine, so it's not much trouble as an external dependency
 - given than vNext will be web-based, creating another Java Swing window feels a bit backwards

But I get your point, it's nice to have the manual in the app. And I wonder if there's any best practice for embedding manuals in modern Java desktop applications: there seems to be no official replacement for javahelp, and the few Java desktop apps I know use online help :confused: .

All in all, I'm not saying the changes in this PR are ideal - for instance, wrt to the in-app manual, there's no table of content so users have to use back and forward browser buttons. However I think it would be better to invest more in improving the manual themselves rather than "the container" (e.g. adding a TOC visible on every page; adding pictures which are in `docs/user/<en/it>/img` which currently are not shown not even on readthedocs for some reason, etc).

lupino3:
I agree 100% that the help needs some love, in terms of structure and navigation. I think a single page might be the best structure for the in-application use case.

To give a bit of context, the main use case for the in-app help was being able to consult the help on lab machines during exams. Those machines would not have any internet connection and could be heterogeneous (Windows / Linux / etc). That's why the help is embedded.

I think your point about the browsers being ubiquitous is valid, but I can also see some scenarios where over-zealous lab assistants may have simply removed browsers to prevent students from browsing during exams :) It's a bit far-fetched but IMHO not impossible.

I'd like to play a bit with JavaFX and see if it's possible without too much effort to embed a JavaFX WebView (which with a bit of effort seems to [support search](https://stackoverflow.com/questions/19418626/javafx-search-and-highlight-text-add-search-bar-for-loaded-web-page)). If that proves too complicated, I think it's fair to rely on the external browser to display the help.

lupino3:
Hey @pviotti, I played a bit with JavaFX and it seems very easy to create a simple viewer. I pushed my initial work here: https://github.com/EduMIPS64/edumips64/pull/497, along with a list of things left to do. The code for the JavaFX window itself is very small, <50 lines of code with all the imports (`HelpViewer.java`).

What do you think about that change?

I would really like to reuse your code to extract the help to a temporary directory, since you've already done the research on reading from the JAR and decided that extracting is the best way.

Could you please extract that helper into a PR which we can merge and that I can use in #497?

pviotti:
> Could you please extract that helper into a PR which we can merge and that I can use in #497?

yep. it's in 681e24310c5354ef375890bcb861f67dbaf604c9 in `feat/decompress-manual` ( `git cherry-pick -n 681e24310` )
(on this branch I just did some reordering and squashing of commits)

> since you've already done the research on reading from the JAR and decided that extracting is the best way.

if it's just a single page should be fine, whereas if you add images I guess it has to be uncompressed (though I haven't tested it)

lupino3:
Hey @pviotti I am closing this PR because I don't think it's necessary to do this change now. This may come in handy if JavaHelp breaks in a worse way in the future. Thanks anyway!

PR #497, opened 2020-10-26
feat: try out possible in-app help viewers

lupino3:
This is very much WIP and temporary, committing to get some feedback.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,12 @@ plugins {
     id ("jacoco")
     id ("com.dorongold.task-tree") version "1.5"
     id ("us.ascendtech.gwt.classic") version "0.5.1"
+    id("org.openjfx.javafxplugin") version "0.0.9"
+}
+
+javafx {
+    version = "15"
+    modules("javafx.base", "javafx.swing", "javafx.controls", "javafx.web")
 }
 
 repositories {
@@ -57,12 +63,12 @@ fun buildDocsCmd(language: String, type: String) : List<String> {
 
 tasks.create<Exec>("htmlDocsEn"){
     workingDir = File("${projectDir}/docs/user/en/src")
-    commandLine(buildDocsCmd("en", "html"))
+    commandLine(buildDocsCmd("en", "singlehtml"))
 }
 
 tasks.create<Exec>("htmlDocsIt") {
     workingDir = File("${projectDir}/docs/user/it/src")
-    commandLine(buildDocsCmd("it", "html"))
+    commandLine(buildDocsCmd("it", "singlehtml"))
 }
 
 tasks.create<Exec>("pdfDocsEn") {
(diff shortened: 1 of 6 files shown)


lupino3:
Here is how it looks like. CSS is rendered correctly:

image

lupino3:
TODO:

- [ ] the panel should resize with the frame
- [ ] all JavaHelp references should be removed
- [ ] the help should be extracted to the filesystem (using @pviotti's code)
- [ ] the help language should match the UI language
- [ ] I think it'd be better to initialize JavaFX at application startup, to avoid delays when loading the help
- [ ] it would be nice to implement a very basic search feature

lupino3:
Ok I think I found a pretty large annoyance with JavaFX: https://github.com/openjfx/javafx-gradle-plugin#4-cross-platform-projects-and-libraries

The runtimes are platform-dependent. And the Gradle tasks by default include runtime libraries for the build platform.

Which in practice means that, as the code stands right now, the EduMIPS64 JAR built on Windows will not run on other platforms such as Mac OS X.

I can see a few ways out, and need to think about the best one:

- provide a JAR for each platform, built via CI. (this will leave out platforms we don't have runners for, like FreeBSD, or might require extra Gradle magic to generate the JARs without running on the given platform);
- create a large JAR with the Windows/Linux/Mac OS X runtimes. The individual platform JARs are already at 38M so it might become too large; 
- try the `JTextPane` approach;

I think I'll have a quick shot at using `JTextPane`, and after that think more about whether it's worth staying on JavaFX or not.

lupino3:
Just for the record, I did a quick and dirty test using `JEditorPane` to display HTML, and..

1. it works
2. it has the same display issue that JavaHelp has:

image

I *suspect* that the encoding issue might be related to something like JavaScript not being run.

lupino3:
Found a way to embed images in `JEditorPane` if we end up taking that route: http://java-sl.com/tip_local_images.html -- a bit hacky but shouldn't be too hard.

pviotti:
> Just for the record, I did a quick and dirty test using JEditorPane to display HTML, and it has the same display issue that JavaHelp has

here (JDK 14, Linux) it works now with the prototype you pushed: the characters encoding is fine, but it badly needs some anti aliasing, and there's some other interesting artifacts...

image

There might be some middle ground between JEditorPane and JavaFX. I just found this, which may be worth looking into https://github.com/flyingsaucerproject/flyingsaucer

lupino3:
Wow, looks terrible! Let me try this flyingsaucer.

lupino3:
FlyingSaucer looks nice, but it has some artifacts on Windows:
image

I think this is the most balanced alternative so far though. Let me see how it looks under Linux/Mac OS X.

I also think that in any case we'll need to tweak the Sphinx theme a bit to get to the best result.

lupino3:
Looks even better on Mac OS X:

Screenshot 2020-10-26 at 20 23 17

lupino3:
FlyingSaucer seems to be a *renderer*, not an *HTML viewer*. By this I mean that it looks like it shows an image of the rendered HTML, which is not selectable and I highly suspect not searchable.

Which might be ok, just something to note.

lupino3:
Closing because no obvious good alternative to JavaHelp emerged from this exploration and in #505 JavaHelp was ultimately fixed.

PR #506, opened 2020-11-08
Chore: find and replace project name in URLs

pviotti:
I noticed the manual was still referencing the old URL so I did a find and replace all over the project (except in a couple of places where it doesn't make sense, e.g. gitter chat).


src/webapp/data/SampleProgram.js
```diff
@@ -25,14 +25,14 @@ result: .word32 0
 	; ---
 
 	.text
-	
+
```
pviotti, line 28: (ah, the white space trimmer in vim, I forgot to disable it -.-)

lupino3: No worries, that's ok :)

lupino3:
Cool, thanks!

PR #616, opened 2021-09-09
Do CI/CD on fixed JDK version and use Zulu instead of Adopt or Temurin

geertjanw:
Adding fixed JDK 11 version for your CI/CD pipeline so that you have a specific version to do setup-java against. If the pipeline against the latest JDK version fails while it continues to succeed against the fixed version, you'll know that the build failure is related to the latest JDK 11 version and not to an error in your code.

Also, since Adopt is no more, the alternative JDK distros to use are Temurin and Zulu, though Temurin doesn't provide all the JDK versions since it is new, while Zulu has a more complete set of JDKs, useful for using fixed versions, so switched from Adopt to Temurin to Zulu.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java }}
-          distribution: 'temurin'
+          distribution: 'zulu'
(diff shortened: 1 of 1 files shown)


lupino3:
Thanks for your contribution! I'm trying to understand why the workflows are failing, and after that I'll remove the required statuses (since their name is changed) and merge your PR.

PR #687, opened 2022-07-01
feat(core): Implement MOV instruction

malbolgee:
This commit implements the MOV instruction. This instruction moves and
immediate value to an resgister.

The syntax for this instruction is: %R, %I - destination register,
immediate.

Signed-off-by: Malbolge <victor.gomes@icomp.ufam.edu.br>

### DIFF
--- a/src/main/java/org/edumips64/core/is/InstructionBuilder.java
+++ b/src/main/java/org/edumips64/core/is/InstructionBuilder.java
@@ -85,6 +85,9 @@ public class InstructionBuilder {
       case "ADDIU":
         instruction = new ADDIU();
         break;
+      case "MOV":
+        instruction = new MOV();
+        break;
 
       //ALU Shifting 32-bits
       case "SLL":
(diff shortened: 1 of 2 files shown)


lupino3:
Hi @malbolgee, thanks for your contribution.

Isn't this instruction equivalent to `daddi %r, r0, %i`?

I'm looking at the MIPS64 ISA (https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MIPS_Architecture_MIPS64_InstructionSet_%20AFP_P_MD00087_06.05.pdf) and I can't find it.

Is there another spec you're using that describes it?

malbolgee:
Hi @lupino3, yes, that's kind of equivalent, I just find this syntax to move the immediate into a register more pleasing than the 'daddi' one. I did not look at any specs that describe it, so I understand if you do not accept this PR. Thank you for the quick analysis.

lupino3:
HI @malbolgee. So, while I'm grateful for your contribution, I'm afraid I cannot accept it.

One of the key principles of the RISC instruction sets is orthogonality, i.e., there should be as few ways of doing one thing as possible. Adding an instruction that is not in any MIPS ISAs and whose purpose is to do something that can be done with another instruction kind of violates orthogonality, so I'd prefer not to include this new instruction you are proposing.

I hope that's ok, and again thanks a lot for taking the time to contribute to EduMIPS64!

PR #710, opened 2022-10-16
Adding SLTUI instruction

galloj:
Instruction is documented here: https://edumips64.readthedocs.io/en/latest/instructions.html but is not implemented. If I am right it should be same as `SLTIU`, so I basically copied how it is done in `DADDUI.java`.

### DIFF
--- a/src/main/java/org/edumips64/core/is/InstructionBuilder.java
+++ b/src/main/java/org/edumips64/core/is/InstructionBuilder.java
@@ -187,6 +187,9 @@ public class InstructionBuilder {
       case "SLTIU":
         instruction = new SLTIU();
         break;
+      case "SLTUI":
+        instruction = new SLTUI();
+        break;
       case "XORI":
         instruction = new XORI();
         break;
--- a/src/main/java/org/edumips64/core/is/SLTUI.java
+++ b/src/main/java/org/edumips64/core/is/SLTUI.java
@@ -1,0 +1,40 @@
+/*
+ * DADDUI.java
+ *
+ * 16th oct 2022
+ * Instruction SLTUI of the MIPS64 Instruction Set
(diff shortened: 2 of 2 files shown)


lupino3:
Hey @galloj, thanks a lot for your contribution!

I believe that what you found is not a missing instruction, but an error in the documentation!

I cannot find an `SLTUI` instruction in the MIPS64 instruction set documentation: https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MIPS_Architecture_MIPS64_InstructionSet_%20AFP_P_MD00087_06.05.pdf

However, there is an `SLTIU` instruction, and that is what is both implemented in code (see [SLTIU.java](https://github.com/EduMIPS64/edumips64/blob/master/src/main/java/org/edumips64/core/is/SLTIU.java) and described in the docs you pointed at.

Since I'd like to preserve a history of your contribution, would you be willing to fix the documentation, replacing the word `SLTUI` with the word `SLTIU`? If you'd prefer, I can also fix it.

Thanks again for finding and reporting this problem!

galloj:
I wasn't really sure if it was missing instruction or error in documentation, but I chose to believe it is missing instruction because both versions of instruction can be found here in your repository: https://github.com/EduMIPS64/edumips64/blob/2b5c97e7df5c0b2a21f7d16f60eba889b5e944b1/contrib/kate_mips.xml

Also when googling for SLTUI instruction, then there are some results (although about 10 times less).

Anyways, it is likely that my guess might be wrong, so if you prefer the fix to documentation, then I will do it.

lupino3:
Yes please! As a reference for the instructions et we follow the MIPS64 official docs I linked earlier, and there is no mention of the `SLTUI` instruction there.

Thank you very much!

PR #733, opened 2022-12-28
Remove gitpod badge

lupino3:


lupino3:
(this PR and the previous one #732 are used to test CI, as I'm moving around the Azure storage account used for integration testing)

lupino3:
> (this PR and the previous one #732 are used to test CI, as I'm moving around the Azure storage account used for integration testing)

Looks like it's fixed now.

PR #737, opened 2023-01-09
chore: upgrade snap base to core22

lupino3:
The recent Sphinx upgrade dropped support for older Python versions, and this broke snap builds. This should fix it.


lupino3:
Cannot be done this way, as the gradle plugin is available at most in `core18`.

PR #738, opened 2023-01-12
chore: update snap Python version

lupino3:


lupino3:
Abandoning in favor of #739

PR #741, opened 2023-01-24
Create merge-dependabot.yml

lupino3:
This code was created by ChatGPT using the prompt "can you give me an example of automatically merging pull requests from dependabot for which all status checks are passing?". Let's see if it works as it is :)

### DIFF
--- a/.github/workflows/merge-dependabot.yml
+++ b/.github/workflows/merge-dependabot.yml
@@ -1,0 +1,27 @@
+name: Dependabot Merge
+
+on:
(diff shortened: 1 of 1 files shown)


lupino3:
Failing with error

```
[Error: .github#L1](https://github.com/EduMIPS64/edumips64/commit/d27b0dff9e5d9d787887bdcc83746b02c9f07296#annotation_8215199901)
every step must define a `uses` or `run` key
```

Asking ChatGPT to fix it

PR #772, opened 2023-09-07
Add Simplified Chinese Support

smallg0at:
This pull request add Simplified Chinese as a language option. 

~~The docs haven't been (and will not be) translated, and will redirect to the English version.~~ ~~Support is work in progress.~~ Translated, but only in-app doc will be available.

This pull request also introduces the following changes:

- Pipeline Window title now properly uses localization strings. 
- Tile MenuItem now have localization strings. 
  - Due to my limited knowledge in Italian, its italian string is currently identical to the English one
 - Add Chinese documentation actions


smallg0at:
Something's going wrong with the code analysis here...

lupino3:
Hi @smallg0at, and thanks for your contribution! I kicked off the CI pipelines, since PR by non-owners by default do not trigger them.

Codacy checks are failing because your changes trigger some tests that were anyway failing: https://app.codacy.com/gh/EduMIPS64/edumips64/pullRequest?prid=12573761. We'll see if I can just bypass it. Let's see if the other build and test pipelines succeed.

I'll be happy to assist you in getting this PR merged and releasing a new version with Simplified Chinese translation, thanks a lot for your work!

src/test/java/org/edumips64/CliRunnerTest.java
```diff
@@ -123,6 +123,14 @@ public void prints_usage_message_it() {
         assertFalse(os.toString().trim().isEmpty());
     }
 
+    @Test
+    public void prints_usage_message_zh() {
+        OutputStream os = getSystemOut();
+        Locale.setDefault(Locale.SCHINESE);
```
lupino3, line 129:
```suggestion
        Locale.setDefault(Locale.SIMPLIFIED_CHINESE);
```

I think this should be `SIMPLIFIED_CHINESE`?
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#SIMPLIFIED_CHINESE

src/test/java/org/edumips64/CliRunnerTest.java
```diff
@@ -153,6 +168,14 @@ public void prints_usage_message_step_it() {
         assertFalse(os.toString().trim().isEmpty());
     }
 
+    @Test
+    public void prints_usage_message_step_zh() {
+        OutputStream os = getSystemOut();
+        Locale.setDefault(Locale.SCHINESE);
```
lupino3, line 174:
```suggestion
        Locale.setDefault(Locale.SIMPLIFIED_CHINESE);
```

Same here, I think.

lupino3 (changes requested):
I think we need to fix one constant to make it compile. I provided suggestions, please accept them if they are correct. Thanks!

smallg0at:
I'm seeing this font fallback issue here

image

lupino3:
> I'm seeing this font fallback issue here. For some reason it doesn't fallback properly, and I am unable to fix :(
> 
> image

I'll have a look at this as soon as possible!

lupino3:
Ok, I think I am able to reproduce the error:

image

FWIW, I had to add the following line to `build.gradle.kts`, otherwise it would not compile on my Windows machine:

image

lupino3:
Ok, I found the problem. We are explicitly setting the Verdana font in `ErrorDialog`, which cannot display Chinese characters, apparently. Removing the explicit font setting fixes the issue:

image

image

This is only a work-around though. We need to manage fonts more consistently in the app.

I'll see if I can think about how to do it tomorrow, so I can unblock your PR, @smallg0at

lupino3:
Hi @smallg0at, I just merged #776, which should fix the errors that caused you problems.

Could you please rebase your changes on top of the new `master` branch?

Then your PR should be good to go.
Let me know if you need any help, and thanks for your contribution!

Once your change is merged, I'll release a new EduMIPS64 version, so you can get the Simplified Chinese version from official packages.

smallg0at:
I'm currently working on translating the docs, so it might take some extra days to complete

lupino3:
Thank you very much, @smallg0at! Appreciate your initiative and effort. Will wait for your work to be finished!

smallg0at:
The translation work is complete, though it still has the following problems:
- Due to the use of machine translation (DeepL), many warnings will appear.
- Chinese docs PDF currently fail to build and will produce empty file

```
         [ERROR] pdfbuilder.py:161 Failed to build doc
         Traceback (most recent call last):
         Traceback (most recent call last):FAILED
           File "D:\Code\edumips64\.gradle\python\Lib\site-packages\rst2pdf\pdfbuilder.py", line 158, in write
             docwriter.write(doctree, destination)
           File "D:\Code\edumips64\.gradle\python\Lib\site-packages\docutils\writers\__init__.py", line 80, in write
             self.translate()
           File "D:\Code\edumips64\.gradle\python\Lib\site-packages\rst2pdf\pdfbuilder.py", line 621, in translate
             contents[0] += nodes.Text(langmod.labels['contents'])
                                       ^^^^^^^^^^^^^^
         AttributeError: module 'zh' has no attribute 'labels'
```

lupino3:
It seems to be building correctly in Github Actions.

smallg0at:
> image
> in-application help works, but it doesn't render well for me

Oops. UTF-8 mysteriously displayed with Windows 1252 encoding. Have no idea what's going on.

> It seems to be building correctly in Github Actions.

Error can be seen [here](https://github.com/EduMIPS64/edumips64/actions/runs/6161815633/job/16753949314#step:10:392)

smallg0at:
Also the actions will need to be updated later for Chinese docs pdf uploading

lupino3:
I fixed the PDF build (we needed to have the right language code for Simplified Chinese in Sphinx), but now the PDF only displays black boxes. Need to investigate a bit more.

lupino3:
We might have a problem with PDF production.. it looks like the library we are using (rst2pdf) may not support simplified chinese output: https://github.com/rst2pdf/rst2pdf/issues/1075. Need to investigate more.

lupino3:
Found a Docker image claiming to be `rst2pdf` set up to output Korean PDF. So I think it is a matter of configuration rather than `rst2pdf` not supporting Chinese fonts.

lupino3:
@smallg0at I tried to copy what's here: https://github.com/S4Plus/ABC/blob/master/toolbox/rst2pdf.md

It still doesn't work for me. Could you please read the page and try to see what I'm doing wrong? Thanks!

lupino3:
It has to be possible, because this PDF looks like it's built with rst2pdf: https://buildmedia.readthedocs.org/media/pdf/test23/1.3.3/test23.pdf

smallg0at:
Now Sarasa Gothic is bundled instead of Simsun to (i) prevent font licensing issues (ii) improve how the manual looks (as Simsun looks terrible on screen). Unfortunately, it still do not work.

smallg0at:
After hours of tinkering I give up. Will try replacing the rst -> pdf chain to [rinohtype](https://github.com/brechtm/rinohtype).

smallg0at:
> After hours of tinkering I give up. Will try replacing the rst -> pdf chain to [rinohtype](https://github.com/brechtm/rinohtype).
> 
> May also allow TeX to return after exiled from this project 12 years ago, if that also didn't work.

Rinohtype attempt failed, and the code can be seen [here](https://github.com/smallg0at/edumips64/tree/rinohtype_attempt).

Chinese PDF exporting support will be dropped for now, and when [this](https://github.com/EduMIPS64/edumips64/pull/772#issuecomment-1717648202) issue got fixed this pull request should be ready for merge.

lupino3:
Thanks, I'll see if I can have a look into the rendering issue today.

lupino3:
https://docs.oracle.com/cd/E19253-01/819-0913/819-0913.pdf Javahelp docs. (for future reference)

lupino3:
Found this tool which has Russian help working with Javahelp, might be able to see how the localization works: https://github.com/logisim-evolution/logisim-evolution.

lupino3:
Just some updates: still not working. I tried using the `SansSerif` font family - since the help was using `Dialog`, with no luck. Also tried adding `xml:lang='zh'` to the help files. Still no luck.

I would suggest merging the PR with the help redirecting to the English manual, just to close the main localization feature, and open an issue for Simplified Chinese in-app help and PDF manual.

What do you think, @smallg0at?

smallg0at:
The help system works for the main audience of this language anyway. I'll check the logisim-evolution implementation and if it doesn't work, put up a warning within the Chinese docs to tell the user what's going on.

smallg0at:
Tested in windows sandbox + English (UK) / Italian (Italy) locale settings. I assume that it's some issue lying in the compile workflow, as no matter what language I'm in the setup always use Chinese. @lupino3 could you trigger the CI action once so that I can see how the GitHub actions build behaves?

smallg0at:
Also this codacy check is ridiculous. Why is this only-repeated-twice string a problem?

lupino3:
> Tested in windows sandbox + English (UK) / Italian (Italy) locale settings, the packages that I built does not have this issue. I assume that it's some issue lying in the compile workflow, as even windows sandbox defaults to english, the setup always use Chinese. @lupino3 could you trigger the CI action once so that I can see how the GitHub actions build behaves?

CI triggered. Note that the language is not inferred from the locale set in environment variables, and instead they are persisted in the user preferences store.

smallg0at:
image

Didn't expect that...

smallg0at:
> > Tested in windows sandbox + English (UK) / Italian (Italy) locale settings, the packages that I built does not have this issue. I assume that it's some issue lying in the compile workflow, as even windows sandbox defaults to english, the setup always use Chinese. @lupino3 could you trigger the CI action once so that I can see how the GitHub actions build behaves?
> 
> CI triggered. Note that the language is not inferred from the locale set in environment variables, and instead they are persisted in the user preferences store.

Actually, Windows sandbox will only inherit the environment locale option (from the physical machine), the user preference language settings will always be en_US initially. The language of the installer does depend on the locale of the computer building it, as seen here:

image

The one on the top is from GitHub release (1.2.10).

lupino3:
Thanks @smallg0at. I had tried the same fix here a few days ago (https://github.com/EduMIPS64/edumips64/pull/777/commits/f32f1ff73f59917bcfe3fdb383deef750c54fe86) but it didn't seem to work :/

smallg0at:
Good news: it has been confirmed fixed.

smallg0at:
There is some hope for rst2pdf however. Logs shows that it cannot locate the stylesheet files, which may be crucial for the implementation. 

image

smallg0at:
Seems to need a token for Webapp distribution to work

https://github.com/smallg0at/edumips64/actions/runs/6237968146/job/16932897585#step:9:23

lupino3:
Thanks for the `read-properties` fix, I'm actually merging it separately here once I see it works: https://github.com/EduMIPS64/edumips64/pull/777 (I'll credit you in the PR)

smallg0at:
image

Found this issue, but in GBK this time. According to my findings this is probably what happened: 

- When changing locale, JDK will automatically change the encoding that it's running on. While it defaults to utf-8 in chinese when installed, the other languages do not get this treatment. HOWEVER, when changing locale to Chinese (China), it will instead modify the settings to be GBK, which adds up another level of mess.
- When loading up a help file, `javax.help.HelpBroker` will NOT respect the HTML encoding and will use the default system-wide java encoding.

smallg0at:
> Found this tool which has Russian help working with Javahelp, might be able to see how the localization works: https://github.com/logisim-evolution/logisim-evolution.

This project use `JHelp` and the implementation can be seen in `src\main\java\com\cburch\logisim\gui\menu\MenuHelp.java`. I can't really code Java so I'm unable to help.

smallg0at:
@lupino3 Let's give it one last try to at least allow the release MSI has normal working Chinese docs. Please set your system variable `JAVA_TOOL_OPTIONS` to `-Dfile.encoding=UTF8`, then build the msi and test it. If it doesn't work, I'll give up and use the English one.

lupino3:
It works!
image
The environment variable was picked up by Gradle correctly, it echoed it during the build process.

I wonder if we can just set it in the Gradle file and have it work for all types of build?

lupino3:
Ok, I verified that it doesn't seem to matter how it's built, but that option needs to be set as an environment variable when the JAR is executed. Let's see if we can work around it.

lupino3:
I'm pretty sure the MSI works because we already pass `-Dfile.encoding=utf-8` to the `java` command line that we build for the Windows shortcut:

image

What we can do is to add this option to the Ubuntu package as well and add a warning at the top of the Chinese docs saying (in English) to load the JAR with `-Dfile.encoding=utf-8` if the help looks garbled.

According to https://stackoverflow.com/questions/361975/setting-the-default-java-character-encoding, there is no way of setting this option within the source code, as it's something used during initialization.

smallg0at:
Thanks for testing. I'll do the following things and then it will be ready to merge:
- [ ] Create Issue: In-app documentation does not respect document encoding
- [ ] Create Issue: CJK characters won't render correctly in the PDF help file
- [ ] Edit docs: add an image in both Chinese and English to state what's happening

smallg0at:
This pull request is ready for merging, as long as Codacy don't do anything funny.

lupino3:
Thanks for your contribution here @smallg0at! I'll see if I can take a look at the issues you filed in the next days, and then I'll create a new release with Simplified Chinese support!

PR #791, opened 2023-09-27
Add: readthedocs configuration

### LINKED ISSUE #788: readrthedocs build failing
https://readthedocs.org/projects/edumips64/builds/21993443/

Seems to use an old Python version (3.7), which is unsupported by our Sphinx version.

smallg0at:
Tries to patch readthedocs generation. Should close #788.

Also build pdf & epub formats, hope nothing goes wrong

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+    # You can also specify other tool versions:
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,35 @@
...
+  tools:
+    python: "3.11"
+    # You can also specify other tool versions:
+    # nodejs: "20"
+    # rust: "1.70"
+    # golang: "1.20"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
```
lupino3, line 19: We don't have this file.. we have a `conf.py` for each language. Try  not setting this option, it looks like if it is set to `null` it will try to find the right file: https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-configuration

lupino3:
Let's see if it works. Thanks!

PR #793, opened 2023-09-29
Change language code name: zh -> zh_CN

smallg0at:


lupino3:
This is not necessary for ReadTheDocs. Do you still want to do this change?

smallg0at:
Probably not, unless someone wants to add Traditional Chinese language.

Issue #779, opened 2023-09-12
DMULU instruction incompatible with WinMips64


smallg0at:
DMULU appears to be implemented as an alias of DMULTU, but the two instructions are actually different.

[WinMips64 Source Code](https://github.com/AndoniZubimendi/WinMIPS64/blob/d1fd07fb4655d2d7d6b47f38f038210a50af301a/src/pipeline.cpp#L63) shows that this instruction actually have 3 register inputs.

It can be tested by loading up factorial.s in WinMips64 examples, which will throw an "invalid register" error.

lupino3:
https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00083-2B-MIPS64INT-AFP-06.01.pdf the MIPS64 ISA confirms that they are two different instructions and should not be aliased. Thanks for the bug report!

lupino3:
Note that if we fix this bug it's worth changing the minor version to 3, as we'd be changing the behavior of one of the instructions (even if the behavior was incorrect).

PR #798, opened 2023-10-08
Fix DMULU and add DMUHU

lupino3:
DMULU was implemented incorrectly as an alias to DMULTU.

This was correct when we first implemented it, as it was a WinMIPS64-specific alias
and not a real MIPS64 instruction. However, the Release 6 of MIPS64 introduced
DMULU with a different semantics.

This change adds it and also implements its complement DMUHU.

As part of this, refactor a bit the tests to add a test harness with parsing
facilities in order to easily test the 32-bit instruction representations.

Also add unit tests, some end-to-end tests and English and Italian documentation
for DMULU and DMUHU.

Thanks @smallg0at for spotting this error and reporting it in #779, which is now fixed.

### DIFF
--- a/docs/user/en/src/instructions.rst
+++ b/docs/user/en/src/instructions.rst
@@ -86,18 +86,23 @@ Here's the list of R-Type ALU Instructions.
   as unsigned values and putting the 32-bits quotient in LO and the 32-bits
   remainder in HI.
 
+* `DMUHU rd, rs, rt`
+
+  Executes the multiplication between 64-bits registers rs and rt,
+  considering them as unsigned values and putting the high-order 64-bits
+  doubleword of the result into register rd.
+
 * `DMULT rs, rt`
 
   Executes the multiplication between 64-bits registers rs and rt, putting
   the low-order 64-bits doubleword of the result into special register LO and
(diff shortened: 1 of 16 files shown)


lupino3:
Note that the web UI test doesn't seem to recognize the new `DMULU` syntax.
Will need to investigate before merging
image

smallg0at:
@lupino3 this might fix the web deploy issue

https://github.com/EduMIPS64/edumips64/pull/802/commits/afd721c3449b64ca130e4aeb78f3c36d74ca8302

lupino3:
> @lupino3 this might fix the web deploy issue
> 
> [afd721c](https://github.com/EduMIPS64/edumips64/commit/afd721c3449b64ca130e4aeb78f3c36d74ca8302)

It might fix it indeed for the PR, but what if the pipeline is run from master? I think we need to specify the right commit in all cases (PR and master)

lupino3:
The web deployment is correct, but the web page loads the master JS code because it does not end with a slash and gets the JS code from the root. Looking at https://edumips64ci.z16.web.core.windows.net/798/ I can find the correct JS code, which supports both `DMULU` and `DMUHU`. Merging the PR

PR #819, opened 2023-10-16
release: 1.3.0

lupino3:


lupino3:
Found an issue: the Italian help doesn't load, and instead the English one is shown.

PR #846, opened 2023-10-23
fix(snap): move to more recent base and build manually

lupino3:


lupino3:
Sample: https://github.com/logisim-evolution/logisim-evolution/blob/b7c028e9fb36b0aa872f0f07cbce51cbbaa94cf1/snap/snapcraft.yaml

PR #1276, opened 2025-03-20
Editor features and new Settings component

davidepatti:
- added load/save code functionality to the main code panel
- added a new Settings component in the sidebar
- implemented vi-mode and font size controls into the Settings section

### DIFF
--- a/src/webapp/components/Code.js
+++ b/src/webapp/components/Code.js
@@ -1,11 +1,9 @@
-
 import React, { useEffect, useState } from 'react';
 import { initVimMode } from 'monaco-vim';
 
 // new
 import MonacoEditor from 'react-monaco-editor';
 import useMediaQuery from '@mui/material/useMediaQuery';
-
 
 const Code = (props) => {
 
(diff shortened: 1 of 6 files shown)


src/webapp/components/Code.js
```diff
@@ -242,18 +291,45 @@ const Code = (props) => {
...
+            borderRadius: '5px',
+            color: 'black',  // 🔥 Fix: Ensures text is black
+            display: 'inline-block',
+            textAlign: 'center'
+          }}>
+            Load Code
+            <input
+                type="file"
+                accept=".txt,.s"
+                onChange={loadCodeFromFile}
```
lupino3, line 317: I think this should only be enabled when the simulator is in editing mode and not when the CPU is running, much like - for example - the 'clear code' button. Otherwise the internal CPU state will be out of sync with the editor.

lupino3 (changes requested):

src/webapp/components/Code.js
```diff
@@ -242,18 +291,45 @@ const Code = (props) => {
     monaco.editor.setModelMarkers(model, 'EduMIPS64', markers);
   }, [props.parsingErrors, editor, monaco]);
 
-  // Set the dark theme if necessary
-  const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
```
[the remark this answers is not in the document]
davidepatti: Solved with the last commit

src/webapp/components/Code.js
davidepatti: I agree. Also, I think "Load/Save Code" stuff should be near to CLEAR CODE, on the top bar of the interface, since they have similar semantics (changing/storing what you are "working on"). If you are ok, I try to move Load/Save on top and add the same "not-when-running" feature.

davidepatti: Moved Open/Save to the header. Similar disabled mode when code is running. 
Screenshot 2025-03-21 at 11 06 19

davidepatti: last commit, made "save code" arrow going the opposite of "open code". Nerds stuff matters.

lupino3 (approved):
Thanks @davidepatti! Overall, looks good to me. I have a few minor tweaks I'd like to do, which can be done later:

1. The Settings accordion looks different, from some reason.. I'll try to make it look the same as the others. 
image

2. I think we could find a way to further differentiate the "Load into simulator" from "Load from file into the editor" actions in the UI.

PR #1278, opened 2025-03-21
Towards a decent Memory UI section

davidepatti:
Just a first improvement, with items separated and organized in colums and decimal values display (see last line)
Suggestions welcome!

Screenshot 2025-03-21 at 11 54 15

### DIFF
--- a/src/webapp/components/Memory.js
+++ b/src/webapp/components/Memory.js
@@ -1,8 +1,67 @@
 import React from 'react';
 
+const parseData = (dataString) => {
+  const rows = dataString.split('\n').map(line => {
+    let address = "", value = "", label = "", code = "", comment = "";
+
+    const addressMatch = line.match(/ADDRESS\s+([^\s,]+)/);
+    if (addressMatch) address = addressMatch[1];
+
+    const valueMatch = line.match(/VALUE\s+([^\s,]+)/);
+    if (valueMatch) value = valueMatch[1];
+
+    const labelMatch = line.match(/LABEL\s+([^\s,]+)/);
+    if (labelMatch) label = labelMatch[1];
+
+    const codeMatch = line.match(/CODE\s+([^,]*)/);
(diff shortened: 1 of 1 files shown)


src/webapp/components/Memory.js
```diff
@@ -1,8 +1,67 @@
 import React from 'react';
 
+const parseData = (dataString) => {
+  const rows = dataString.split('\n').map(line => {
```
lupino3, line 4: Regardless of the UI presentation, my intention here was to construct in the Java code a proper array of JSON objects containing the information that the UI can display, which can then be manipulated in the UI more easily than using regex to extract bits of information.

I think doing so is also nice, because it'll show you how I pass data from Java to JS.

The Registers section does this currently, here is the relevant code:

* code to read Java Registers and build JSON: https://github.com/EduMIPS64/edumips64/blob/49e284995b01a1a97bfdf226364a643ff414d395/src/main/java/org/edumips64/client/ResultFactory.java#L108
* code that updates the internal JS state with the new registers: https://github.com/EduMIPS64/edumips64/blob/master/src/webapp/components/Simulator.js#L89
* the Register React component: https://github.com/EduMIPS64/edumips64/blob/master/src/webapp/components/Registers.js#L22

For the Memory cell, you could also reuse the same `BinaryValue` component that is in `Registers.js`, moving it to a new file.

lupino3 (changes requested):

src/webapp/components/Memory.js
davidepatti: of course, agree with you. But in the meanwhile, why keep that unreadable mess and don't just merge this clean formatted fix?

davidepatti: ok @lupino3 , I'll try to proceed as suggested while learning how to deal with that java->js stuff, stay tuned

lupino3:
Closing in favor of #1289

Issue #1225, opened 2025-02-09
Web UI: improve memory widget


lupino3:
Add individual cells, ability to inspect contents, convert from hexadecimal to decimal.

lupino3:
Fixed by #1289. Thanks @davidepatti!

PR #1289, opened 2025-03-28
JSON-based WebUI Memory Section

davidepatti:
This contribution replaces [previous wrong approach](https://github.com/EduMIPS64/edumips64/pull/1278)
Data are exported from Java with JSON object memory containing arrays for cells and intructions.
Memory.js renders the attributes exported form ResultsFactory.
Please notice that a couple of methods were added to extract some values from Memory java component, it seemed necessary, but I'm not 100% confident.

Suggestions are welcome.
Screenshot 2025-03-28 at 10 30 47


lupino3:
Will fix #1225

src/webapp/components/Memory.js
```diff
@@ -1,9 +1,68 @@
...
+                    <th style={{textAlign: 'left'}}>Comment</th>
+                </tr>
+                </thead>
+                <tbody>
+                {cells.map((memoryelement, i) => (
+                    <tr key={i}>
+                        <MemoryElement memoryelement={cells[i]}/>
+                    </tr>
+                ))}
+                {instructions.map((memoryelement, i) => (
```
lupino3, line 57: I think I would not display the code memory here, as we are already showing the code in the editor and I find it confusing that here we show both data and code memory.

How about we do not show instructions here, and instead (maybe in another PR) we figure out a way to show memory addresses in the editor itself? (maybe via custom line numbers?)

src/webapp/components/Memory.js
```diff
@@ -1,9 +1,68 @@
...
+          textAlign: 'right',
+        }}>
+          {hexString}
+        </Typography>
+      </Tooltip>
   );
 };
 
+
+const MemoryElement = ({memoryelement}) => {
```
lupino3, line 23: Have you considered using `<BinaryValue>` for the address and the value? It will auto-convert hex to dec in the tooltip. It is currently buried inside `Registers.js`, so it needs to be extracted.

src/webapp/components/Memory.js
```diff
@@ -1,9 +1,68 @@
...
+
+const Memory = (props) => {
+    const memory = props.memory;
+    const {cells = [], instructions = []} = memory;
+    return (
+        <div>
+            <table id="memory">
+                <thead>
+                <tr>
+                    <th style={{textAlign: 'left'}}>Address</th>
```
lupino3, line 44: Should we show something like "no program loaded" rather than these headers when the simulator is not running?

lupino3 (commented):
The overall approach looks good, thanks! I have a few comments and I think we're done for a v1

src/webapp/components/Memory.js
davidepatti: Yes! It absolutely makes sense to me. That "show instruction memory" behaviour was a legacy from the swing UI, due to a missing editor. And having now an editor that shows pipeline stages on the code makes it useless waste of space. I'll remove it.

davidepatti: Removed instructions in the last commit, fixing other reviews will come later. Also, I left unchanged all the java side stuff for instructions, e.g., in ResultsFactory, maybe it will useful for the future pull request on code editor memory address display? I can figure out at the moment, keeping a conservative approach.

src/webapp/components/Memory.js
davidepatti: Fixed

src/webapp/components/Memory.js
davidepatti: It would ok but (at least for me) it has an insane ratio of "improvement_you_get"/"things_to_deal_with".  I tried to look around, to import the status of the not loaded code from the event of clicking "Reset code" OR from just having some code open with the new "open code" button, but non having it loaded... I will be honest, I will not spend time on this at the moment :)

davidepatti: Ok, managed directly into the Memory rendering logic.

lupino3:
@davidepatti the auto-preview now works! After I approve the deployment, you can now see your version of the Web UI when you click on "View Deployment"

image

src/main/java/org/edumips64/client/ResultFactory.java
```diff
@@ -102,6 +102,43 @@ private Result AddCpuInfo(Result r) {
     }
 
     private String getMemory() {
+        var memory = new FluentJsonObject();
+        try {
+            var cells = cpu.getMemory().getCells();
```
lupino3, line 107: Now that I think about it, this object already has a reference to `memory`! You can use that one.

Could you also please remove `CPU.getMemory()`? I think it should never be needed.

lupino3: Note that you may not be seeing it because in line 105 you create another variable called `memory`, therefore hiding `this.memory`. You could call the JSON object `memoryJson` and you should be able to just access `this.memory` by calling it `memory`.

lupino3 (approved):
Overall it looks good, thanks! Please address the comment about accessing `memory` and we're good to go!

src/main/java/org/edumips64/client/ResultFactory.java
davidepatti: Done

lupino3:
Fixes #344

PR #1296, opened 2025-03-30
DOES NOT WORK. Non-working stdin implementation.

lupino3:
This is how a naive web-based stdin implementation would look like, except that the web worker doesn't have a way of "waiting" for the data to come from the UI.

A better solution would probably be event-based,
but I am unsure if it can be implemented without
changing more deeply the simulator logic and the way it interacts with I/O.


lupino3:
@davidepatti suggests to implement a new CPU state to represent waiting for the input, which makes total sense.

PR #1299, opened 2025-04-01
Remove default expansion from StdOut's Accordion

davidepatti:
StdOut section should be expanded only when inspected by user, coherently with Registers, Pipeline, Memory.
Also, many programs just don't produce any stdout and keeping it expanded with a blank box consumes a lot of space on the right side, making settings tab not visible in many resolutions. Hopefully, Settings tab will be filled with more controls that user wants to access easily.

### DIFF
--- a/src/webapp/components/Simulator.js
+++ b/src/webapp/components/Simulator.js
@@ -325,7 +325,7 @@ const Simulator = ({ sim, initialState, appInsights }) => {
                 <Memory memory={memory} />
               </AccordionDetails>
             </Accordion>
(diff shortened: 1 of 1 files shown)


lupino3 (approved):
Looks good, thanks! Not merging it yet because for some reason the CI-deployed version does *not* show your change, but I confirmed that if I build the code in your PR the behavior is the one you explain. I'd like to debug it today/tomorrow and then I'll merge the PR. Thanks

lupino3:
Filed #1301 with the CI bug, will merge this PR now.

PR #1306, opened 2025-04-04
Fix bit width in `getValueHex` method.

davidepatti:
For some reason, probably as legacy of the old 32/64 bit lenght distinction for memory section code/data, WebUI memory is initialized using hex values computed using `positiveIntToBin` on 32 bits.
The resulting effect is shown in the first pic, with the second column of different sizes.
This could be acceptable for the first column, like "ok addresses are 64 bits but they are a read-only field and we gain interface space since we will never need so distant memory regions."
But for the second column generate confusing effects, as shown in the second pic where storing the byte "6" on location 50 casuses the line to be filled with 0's, but only on the right.
This PR put a 64 bit length for hex value since the beginning to avoid this weird behaviours (see last pic),

**Memory with weird lengths:**
mem_32bit
**before storing the byte:**
store_byte_pre
**after storing the byte:**
store_byte
**behaviour with this PR:**
64_store_byte

### DIFF
--- a/src/main/java/org/edumips64/core/MemoryElement.java
+++ b/src/main/java/org/edumips64/core/MemoryElement.java
@@ -117,7 +117,7 @@ public class MemoryElement extends BitSet64 {
   }
   public String getValueHex() {
     try {
-      return Converter.binToHex(Converter.positiveIntToBin(32, this.getValue()));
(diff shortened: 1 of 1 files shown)


lupino3:
Thank you!

PR #1309, opened 2025-04-04
Embedded Cache Simulator

davidepatti:
This is an initial DRAFT for embedding a Dinero-like cache simulator inside Edumips64 java code.
The current component CacheSimulator is compatible with Dinero xdin trace files already produced by Edumips64.

Design goals:
- Provide a self contained cache simulator that focused on the 3 main paramters of a cache: size, block size and associativity, using DineroIV default parameters for all the other stuff (Replacement policy LRU,  Write allocate policy always)
- Allowing the WebUI and the core edumips a chance to be independent from the unmantained 2002 code of Dinero, providing a direct integration on the whole project
- Still Keeping the Swing UI dinero frontend for using external dinero if anyone wants to use it

TODO: 
- test to against DineroIV defaults (all positive so far)
- UI interface elements to configure cache parameters and see results.

Stay tuned!


davidepatti:
Further notes before the next updates:
- I think DineroIV compatibility should be maintained because some people just want to use edumips64 and provide a trace to the best well-known cache simulator.
- Neverheless, Edumips64 was born at university, and its main goals are research and education, thus an integrated component offers a seamless way to experiment.... do you have a great idea for a new cache algorithm? just test it on edumips!

lupino3:
There is an old cache implementation here if you want to take a look: https://github.com/EduMIPS64/edumips64/tree/cache. I never had time to integrate it.

davidepatti:
> There is an old cache implementation here if you want to take a look: https://github.com/EduMIPS64/edumips64/tree/cache. I never had time to integrate it.

It seems to be a different and MORE complex approach, i.e., putting an actual memory implementation in edumips that interacts with the other elements. It actually puts and mve the real data inside the memory cells.... I wanted to achieve another goal: you already have this trace file, then you can test on a cache configuration. I don't want to see "what's inside" the memory, I don't want to store the actual data inside the cells. The statistics of a cache are intependent from its content. I hope this makes sense.

davidepatti:
First alpha version:

- The Dinero component, originally just a small class collecting memory references to be processed later,  has been generalized to a larger CacheSimulator component that dynamically receives memory accesses request
- While the program is running, Cache stats are dynamically updates, same as it happens with other things (cycles, RAW stall etc..) (see pic)
- A new cache config sections has been added to set the main cache parameters from WebUI
- Notice that the processTrace of CacheSimulator could be replaced with a any other implementation. However seems very similaro to dineroIV behaviour.

Possible improvements:
- add the Java GUI settings (should be triviial)
- some more tests & suggestions!
Screenshot 2025-04-08 at 16 50 38

davidepatti:
With the last two commits this should be the final version (when considering features). 
Including Java GUI features in this PR so that things are managed coherently.

lupino3:
@davidepatti the approach looks good overall, but could you please add some unit tests?

For example, adding some dedicated cache-only tests in https://github.com/EduMIPS64/edumips64/blob/master/src/test/java/org/edumips64/EndToEndTests.java, and adding to some of the e2e tests some checks for the new statistics as well.

The other missing thing would be updating the manual for this new feature, but we can do it in a separate PR.

lupino3:
@davidepatti I see a problem here. I run this program through both the Web UI and the Java UI:

```mips
.code

ADDI r1, r0, 1
ADDI r2, r0, 1
ADDI r4, r0, 2
DADDU r3, r1, r2
BEQ r4, r3, exit
BREAK

exit: SYSCALL 0
```
The Web UI behaves well
image

While the Java UI has a very strange behavior, and the reads go up and down, finishing at 0
image

Can you please check what's wrong?

src/main/java/org/edumips64/ui/swing/GUIStatistics.java
```diff
@@ -188,6 +201,28 @@ public Component getListCellRendererComponent(JList<? extends String> list, Stri
       case 15:
         label.setText(" " + cpu.getFCSR().getBinString());
         return label;
+      case 16:
+        label.setText(" " + CurrentLocale.getString("L1CACHESTATS"));
```
lupino3, line 205: Can you please add all these strings you use in GUIStatistics in CurrentLocale? They show up as they are right now and it is not easy to understand what they are about.

src/main/java/org/edumips64/core/CacheSimulator.java
```diff
@@ -0,0 +1,384 @@
...
+    final private CacheMemory L1InstructionCache;
+    final private CacheMemory L1DataCache;
+
+    private LinkedList<String> dineroData = new LinkedList<>();
+    // Offset of the data segment. This class writes a trace file that assumes
+    // that the data segment starts immediately after the code segment ends.
+    private int offset;
+
+    public CacheSimulator() {
+        // fake values to be replaced
```
lupino3, line 21: Are these defaults? Do we want to load them from the configuration?

src/main/java/org/edumips64/core/CacheSimulator.java
```diff
@@ -0,0 +1,384 @@
...
+        public String toString() {
+            return "CacheMemory{" + type + config + '}';
+        }
+
+        public Stats getStats() {
+            return stats;
+        }
+    }
+
+    public static void main(String[] args) {
```
lupino3, line 144: Please remove this main()

src/main/java/org/edumips64/ui/swing/GUIComponent.java
```diff
@@ -39,15 +40,17 @@ abstract class GUIComponent {
...
   Font font;
   float scalingDenominator = 12.0f;
 
-  GUIComponent(CPU cpu, Memory memory, ConfigStore config) {
+  GUIComponent(CPU cpu, Memory memory, CacheSimulator cacheSimulator, ConfigStore config) {
     cont = null;
     this.cpu = cpu;
     this.memory = memory;
     this.config = config;
+    this.cachesim = cacheSimulator;
```
lupino3, line 53: Is this needed by all components? I'd imagine only GUIConfig and GUIStatistics truly need it.

src/test/java/org/edumips64/CacheSimulatorTests.java
```diff
@@ -0,0 +1,189 @@
...
+                cache.stats.getWriteAccesses(),
+                cache.stats.getWriteMisses()
+        );
+    }
+
+    @Before
+    public void setUp() {
+    }
+
+    public static void main(String[] args) {
```
lupino3, line 99: Can you remove this?

src/test/java/org/edumips64/EndToEndTests.java
```diff
@@ -211,7 +207,7 @@ private CpuTestStatus runMipsTest(String testPath, boolean writeTracefile) throw
       return new CpuTestStatus(cpu, tracefile);
     } finally {
       cpu.reset();
-      dinero.reset();
+      //cachesim.reset();
```
lupino3, line 210: Why is this commented?

src/webapp/components/CacheConfig.js
```diff
@@ -0,0 +1,64 @@
...
+                <div>
+                    <Typography sx={labelStyle}>Size</Typography>
+                    <input type="number" size="4" value={l1d.size} onChange={e => setL1D({ ...l1d, size: parseInt(e.target.value) })} disabled={status === 'RUNNING'} />
+                </div>
+                <div>
+                    <Typography sx={labelStyle}>Block Size</Typography>
+                    <input type="number" size="4" value={l1d.blockSize} onChange={e => setL1D({ ...l1d, blockSize: parseInt(e.target.value) })} disabled={status === 'RUNNING'} />
+                </div>
+                <div>
+                    <Typography sx={labelStyle}>Associativity</Typography>
```
lupino3, line 32: Associativity and Penalty overflow in my browser.. can you please fix it?

image

lupino3 (changes requested):
Ciao Davide, overall this looks good. I added a few comments to the PR, there are a couple of things that I think need to be fixed. I also opened a separate issue for the manual ;) https://github.com/EduMIPS64/edumips64/issues/1321

src/main/java/org/edumips64/ui/swing/GUIStatistics.java
davidepatti: Fixed [here](https://github.com/EduMIPS64/edumips64/pull/1309/commits/fed5b82f7d5e18eeca17f68eb951d70ce0ae4151)

src/main/java/org/edumips64/core/CacheSimulator.java
davidepatti: Fixed [here](https://github.com/EduMIPS64/edumips64/pull/1309/commits/0005bfad053b5725fa58309eb22bbe9d91f3e84f)

src/main/java/org/edumips64/core/CacheSimulator.java
davidepatti: fixed

src/main/java/org/edumips64/ui/swing/GUIComponent.java
davidepatti: Fixed [here]( https://github.com/EduMIPS64/edumips64/pull/1309/commits/1fe51b406686a0768d2cd405a487ff99795f3009)

src/test/java/org/edumips64/CacheSimulatorTests.java
davidepatti: fixed

src/test/java/org/edumips64/EndToEndTests.java
davidepatti: Resulting status is required (reset done when required, outside). Removed comment.

src/webapp/components/CacheConfig.js
davidepatti: Hopefully fixed [here](https://github.com/EduMIPS64/edumips64/pull/1309/commits/0005bfad053b5725fa58309eb22bbe9d91f3e84f)

davidepatti:
> @davidepatti I see a problem here. I run this program through both the Web UI and the Java UI:
The problem with the not updated cache stats in java version was related to CPUSwingWorker calling setConfig at each cycle, resetting the caches. This is why it not happened when "running all". The resetting logic handlig has been fixed [here](https://github.com/EduMIPS64/edumips64/pull/1309/commits/7c34389b9b14556cfbcdaab1d6fe3fda89063169)

src/main/java/org/edumips64/core/CacheSimulator.java
```diff
@@ -112,7 +111,8 @@ public void setConfig(ConfigStore config) {
             setConfig(cc);
         }
 
-        public void resetStats() {
+        public void resetStatus() {
+            setConfig(this.config);
```
lupino3, line 126: I don't understand this change. Why do you call `setConfig`? The configuration is exactly the same, and also `setConfig` has the side effect of calling `this.stats.reset()`, so actually you will reset the stats object twice.. is it intended? Shouldn't `resetStats()` just call `this.stats.reset()`?

src/main/java/org/edumips64/utils/CurrentLocale.java
```diff
@@ -148,6 +148,13 @@ public static void setConfig(ConfigStore config) {
     en.put("BTS", "Branch Taken Stalls");
     en.put("BMS", "Branch Misprediction Stalls");
     en.put("CSIZE", "Code size");
+    en.put("L1CACHESTATS", "Cache Memory Statistics");
```
lupino3, line 151: I won't ask for Chinese :D but can you please add Italian as well?

src/main/java/org/edumips64/core/CacheSimulator.java
lupino3: I may be missing something, but how was it fixed? I still see the comment "fake values to be replaced", and some numbers I don't fully understand.

src/main/java/org/edumips64/core/CacheSimulator.java
lupino3: The `main` method is still here, or am I crazy?

src/main/java/org/edumips64/ui/swing/GUIComponent.java
```diff
@@ -24,6 +24,7 @@
 package org.edumips64.ui.swing;
 
 import org.edumips64.core.CPU;
+import org.edumips64.core.CacheSimulator;
```
lupino3, line 27: This can be removed now

src/main/java/org/edumips64/ui/swing/GUIComponent.java
lupino3: Thanks! Feel free to "Resolve" conversations in github once you implement my suggestions, and leave them open if you want to discuss things more. Thanks!

src/test/java/org/edumips64/CacheSimulatorTests.java
lupino3: nope, it's still here!

src/webapp/components/CacheConfig.js
lupino3: It is much better, but still breaks when the browser window is made smaller. No worries, I created #1331 to fix it later.

src/main/java/org/edumips64/core/CacheSimulator.java
davidepatti: ehmmm.... sorry!

src/test/java/org/edumips64/CacheSimulatorTests.java
davidepatti: ehmmm.... sorry bis!

src/main/java/org/edumips64/core/CacheSimulator.java
davidepatti: ok, was a lazy way to empty the cache content.... now we have a separate method eraseCacheContent so the we don't abuse of setConfig....

src/main/java/org/edumips64/utils/CurrentLocale.java
davidepatti: maybe sicilian in some next release ?

davidepatti: (I did it for italian)

lupino3 (approved):
Thanks for the changes! Looks good! I'll resolve conflicts and merge.

src/webapp/components/Simulator.js
```diff
@@ -170,6 +171,10 @@ const Simulator = ({worker, initialState, appInsights}) => {
   }
 
 
+  const setCacheConfig = (config) => {
+    sim.setCacheConfig(config);
```
lupino3, line 175:
```suggestion
    worker.setCacheConfig(config);
```

Issue #1317, opened 2025-04-16
Wrong parsing of hexadecimals in format I instructions


davidepatti:
Hexadecimals in Format I instructions are signed 16 bit values.
This means that something like:

`daddi r1,r0,0xffff`

should not overflow, because it's sixteen bits equal to 1, which will be sign extended to 64 bits, and  it's just -1.

Instead, the following confusing thing is allowed:

`daddi r1,r0,0x-0001`

The problem arises from using this code in Converter.java

```
    if (immediate.length() >= 3 && immediate.substring(0, 2).compareToIgnoreCase("0x") == 0) {
      immediate = immediate.substring(2);
      base = 16;
    }
    return Long.parseLong(immediate, base);

```

Thus, the string gets the "0x" prefix removed and then parseLong converts as 16 base number, including a sign.
I think the most useful and commond purpose of 0x hex expressions is to describe the bit content of a field rather then just changing the base. 
For example, when you look at the result in R1, you will see FFFFFFFFFFFFFFFF, and not -000000000000001

Trying to write a PR request that keeps everything unchanged, just handling the case when you have an hex number is the range 32768-65535, that now is considered overflow, while should be just converted to its signed correspondent number (just by subtracting 65536)

davidepatti:
Anyway, hex immediate values seems to have various problems that maybe should addressed once for all:

Originally, there seems to have been some problems solved [here](https://github.com/EduMIPS64/edumips64/pull/608)

Anyway, I already mentioned a problem with signed 16 bit immediate (see above), so that a signed 16-bit 0xffff is not -1 but overflow...

But in the meanwhile, I see stuff like:

`label: .word 0xdeadfffffffffffffffffffffffffffffff`

 accepted by the parser,  just truncating, without any complain or warning.

davidepatti:
Fixed in [this PR](https://github.com/EduMIPS64/edumips64/pull/1318)

PR #1318, opened 2025-04-22
Fixed hex parsing

davidepatti:
This PR solves the following:

- In .data section, hex strings unreasonable long hex strings are checked before any conversion. For example, there's no chance of trying to apply conversion to something that is > 18 chars trimmed (8 bytes + 0x).
- In .data section, each hex string is checked against its actual size  (byte, word16 etc...) , not according some converted value. You should be able to write .byte 0xff to put 8 ones somewhere.
- In code, now is possible to use all the 16 bits of immediate in hex strings
- The mem section in webUI suffered a little bug in displaying big hex values, due to unmotivated usage of a different method rather the re-using the same value used in the Java version.

### DIFF
--- a/src/main/java/org/edumips64/client/ResultFactory.java
+++ b/src/main/java/org/edumips64/client/ResultFactory.java
@@ -30,6 +30,7 @@ import java.util.logging.Logger;
 import com.google.gwt.json.client.JSONArray;
 
 import org.edumips64.core.CPU;
+import org.edumips64.core.IrregularStringOfBitsException;
 import org.edumips64.core.Memory;
 import org.edumips64.core.Register;
 import org.edumips64.core.CPU.CPUStatus;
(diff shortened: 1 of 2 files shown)


davidepatti:
Added test for these special cases handling. All the other pre-existing tests working as usual, which confirms that these edge cases were just not handled and tested.

src/test/java/org/edumips64/core/ParserTest.java
```diff
@@ -271,4 +271,23 @@ public void MissingParmeterTest() throws Exception {
...
     parseCode("LW r1");
   }
+
+  @Test(expected = ParserMultiException.class)
+  public void unlimitedHexStringTest() throws Exception {
+    parseData(".word 0xdeadbeeffffffffffffffffff");
+  }
+
+  @Test
+  public void hexStringLimitData() throws Exception {
```
lupino3, line 281: Note that this test will "succeed" if any of the below `parseData` calls raises an exception. If you expect all of them to raise exceptions, they need to be separate tests.

src/test/java/org/edumips64/core/ParserTest.java
```diff
@@ -271,4 +271,23 @@ public void MissingParmeterTest() throws Exception {
     // Fully missing parameter. Before the fix for #1048, this would throw StringIndexOutOfBoundsException.
     parseCode("LW r1");
   }
+
```
lupino3, line 274: Please add a comment explaining that the below are regression tests for #1317

src/main/java/org/edumips64/client/ResultFactory.java
```diff
@@ -109,15 +110,19 @@ private String getMemory() {
 
             JSONArray cellArray = new JSONArray();
             cells.forEach((address, element) -> {
-                cellArray.set(cellArray.size(), new FluentJsonObject()
-                        .put("address_hex", element.getAddressHex())
-                        .put("address", element.getAddress())
-                        .put("value", element.getValue())
-                        .put("value_hex", element.getValueHex())
```
lupino3, line 116: Do we still use/need `getValueHex()`? If not, can you please delete it?

lupino3 (changes requested):
Looks good! Just a couple of minor comments.

src/test/java/org/edumips64/core/ParserTest.java
davidepatti: This test does not expect any execption. To succeed, alla the parseData must be ok.

lupino3: Right, sorry!

lupino3 (changes requested):
@davidepatti there are 2 edge cases not covered by unit tests (see Codecov patch comments on `Parser.java`). If you have time, could you add 2 unit tests to cover them? I feel they are important enough that they need to be represented in `ParserTest.java`. Other than that, the PR is good to go. Thanks!

lupino3:
Thanks!

PR #1653, opened 2026-04-20
Await code

tejas-ae:
While tracing code in src/webapp/components/HelpDialog.js I noticed what might be an issue: The promise result gets used as if it were already resolved. this patch adds the missing await so the async path is actually sequenced. Let me know if the approach doesn’t fit.
Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.

### DIFF
--- a/src/webapp/components/HelpDialog.js
+++ b/src/webapp/components/HelpDialog.js
@@ -148,7 +148,7 @@ function useToc(language) {
       }
     }
 
-    fetchToc();
+    await fetchToc();
 
     return () => {
       abortController.abort();


lupino3:
Hey @tejasae-afk, first of all thank you for your contribution, appreciate it!

Based on my understanding, you cannot `await fetchToc()` because `useEffect` is not async, and `fetchToc()` must be async because `fetch` must be awaited. I think the original code is correct. 

Have you tried building your code locally? Based on what I see on CI it fails to compile.

PR #1675, opened 2026-04-21
Add WebUI setting for toggling forwarding in simulation

davidepatti:
### Description
The code for forwarding was already in the main simulation engine, only its wiring to WebUI was missing.
This PR adds an Enable Forwarding checkbox to the web app’s General Settings and wires it to the simulator backend so it behaves like the Java client.

Changing the setting now:

- prompts the user for confirmation,
- resets the simulator,
- reloads the currently loaded program with the new forwarding mode,
- keeps the UI checkbox synchronized with the actual simulator state.

The PR also adds coverage for the new behavior in both the Java-side worker path and the web UI flow.
Validation
```
npm run build
./gradlew test --tests org.edumips64.client.ResultTest --tests org.edumips64.EndToEndTests.testForwarding
```

Playwright test for the new setting flow in src/test/webapp/forwarding-setting.spec.js


lupino3:
Davide -- sorry but I didn't see this and told the robots to do the very same thing... #1676 was just merged. Apologies :( 

I think I will keep the approach of my PR, as I also added unit tests to verify cycle numbers with and without forwarding.

davidepatti:
> Davide -- sorry but I didn't see this and told the robots to do the very same thing... #1676 was just merged. Apologies :(
> 
> I think I will keep the approach of my PR, as I also added unit tests to verify cycle numbers with and without forwarding.

No worries!
I think our agents should talk each other more often :)

PR #1724, opened 2026-05-01
Fix #1723: stop reinstalling the Monaco tokenizer on every render

### LINKED ISSUE #1723: Web UI: instruction/comment text becomes unreadable on row sitting in matching pipeline stage (color collision)
# Web UI: instruction/comment text becomes unreadable when row sits in matching pipeline stage (color collision)

## Summary

While a program is executing in the web UI, instruction mnemonics on the row currently in the **ID** stage become effectively invisible, and comments on the row currently in the **MEM** stage become effectively invisible. The user perceives this as "syntax highlighting disappears during execution".

The root cause is a **color collision** between the per-line stage decoration backgrounds (50%-alpha primary colors) and the Monaco syntax token foreground colors (full-saturation primary colors), both defined in the web UI.

## Reproduction

1. Open https://web.edumips.org with the default sample program loaded.
2. Click **Load**.
3. Click **Run All** (or **Single Step** repeatedly while watching an instruction line). At any frame where:
   - an instruction row is in the ID stage → its blue mnemonics (`daddi`, `bne`, `sw`, ...) blend into the blue ID-stage row background.
   - a comment row is in the MEM stage → its green text blends into the green MEM-stage row background.

This is most visible during **Run All** because the camera passes through every instruction and every comment line that the PC walks past.

## Evidence

DOM-level confirmation (current production build):

| Token | Computed text color | Stage class on row | Computed row background |
|---|---|---|---|
| `daddi`, `bne`, `sw`, `sd`, `syscall` (mtk6) | `rgb(0, 0, 255)` | `.stageId` | `rgba(0, 0, 255, 0.5)` |
| comment text (mtk8) | `rgb(0, 128, 0)` | `.stageMem` | `rgba(0, 255, 0, 0.5)` |
| register `r0..r31` (mtk20, dark red) | `rgb(128, 0, 0)` | `.stageEx` | `rgba(255, 0, 0, 0.5)` (mild collision) |

The colors are defined in `src/webapp/static/style.css` (lines 102–133) and the mappings to Monarch token types are defined in `src/webapp/components/Code.js` (lines 28–42) — note that `mtk6`/`mtk8`/etc. are Monaco's auto-assigned classes derived from the default vs / vs-dark theme palette, not authored explicitly. The collision is therefore an unintended interaction.

## Affected files

- `src/webapp/static/style.css` — stage decoration background colors
- `src/webapp/components/Code.js` — Monarch tokenizer token type assignments and `deltaDecorations` calls (lines 56–161)

## Proposed fix

Change the stage decoration backgrounds so that no full-line background equals (or strongly tints toward) the foreground color of any token routinely placed on that line. Options:

1. **Lower alpha + scope under .monaco-editor** so tokens stay readable. E.g. `rgba(0, 0, 255, 0.18)` instead of `0.5`. Cheap, preserves the color identity matching the Pipeline panel cells.
2. **Use border-left or text-decoration markers** instead of full-row backgrounds, leaving the editor canvas untouched.
3. **Re-pick stage colors from a palette that doesn't intersect Monaco token colors** — e.g. pastel hues that avoid pure blue and pure green channels.

(1) is the smallest diff and preserves the Pipeline panel ↔ editor row visual link. Recommended.

## Acceptance criteria

- [ ] During Run All (default sample), all instruction mnemonics and comment text remain readable on every row, regardless of which stage decoration is applied.
- [ ] The Pipeline panel cells and Code editor row decorations remain visually associated (same hue family).
- [ ] No regression in light or dark Monaco theme.

## Related

- #1697 (panel defaults during execution) — different bug, but also affects perceived "is the runtime working?"
- #1701 (Pipeline visualization relies on color alone) — related a11y theme; this fix should remain colorblind-safe-as-before.

lupino3:
## Summary

Fixes #1723 for real this time.

### What the bug actually is

While a program is executing, **all** syntax highlighting in the Monaco editor disappears: instruction mnemonics, register names, comments, strings, directives — everything renders as plain black text. As soon as the simulator stops, the colours come back.

### Root cause

`Code.js` was calling `monaco.languages.setMonarchTokensProvider('mips', …)` directly in the component body, so the tokenizer was reinstalled on every render of `Code`. During Run All, the parent component re-renders many times per second (one per pipeline-state update), which means `setMonarchTokensProvider` was being called continuously throughout execution.

Each call to `setMonarchTokensProvider`:

1. invalidates the current tokenization for every line of the model, and
2. kicks off an async re-tokenization pass.

While that async pass is in flight Monaco renders the affected lines using the default token class (`mtk1`, `rgb(0, 0, 0)`). Because the calls are happening continuously the editor never gets a chance to finish a tokenization pass before the next reinstall arrives, so the editor stays "black" for the entire duration of the run.

I confirmed this by hooking `setMonarchTokensProvider` on the live preview build and counting calls per second:

```
after Load:        3 calls
during Run All:  ~30 calls/second
total over run:  171 calls
```

…and by sampling the editor DOM mid-run: every span on every `.code` line carried `class="mtk1"` and `getComputedStyle(span).color === "rgb(0, 0, 0)"`.

### The fix

Move the `setMonarchTokensProvider` call into a `useEffect` keyed on `props.validInstructions`. The instruction set never changes after the first worker hand-off (and across hot reloads in dev), so the provider is now installed exactly once for the lifetime of a session instead of dozens of times per second during execution.

### Why the previous commit on this branch did not fix the bug

I previously force-pushed a commit that lowered the alpha of the pipeline-stage decoration backgrounds in `style.css`, on the theory that the `.stageId` blue overlay was masking the blue keyword colour. That diagnosis was wrong: the underlying tokens were not actually blue during execution, they were already `mtk1`/black, so reducing the decoration alpha had no effect on the symptom the user reported. That commit has been replaced via force-push with the real fix in `Code.js`. No CSS changes are part of this PR anymore.

### Regression test

`src/test/webapp/syntax-highlighting-during-run.spec.js`:

- loads the default sample program,
- starts Run All,
- polls the editor DOM repeatedly during the run,
- on each poll, looks at every visible `.view-line` whose text contains a known mnemonic (`daddi`, `bne`, `nop`, `syscall`, …) and checks whether *every* `mtk*` span on that line is `rgb(0, 0, 0)`,
- fails if more than one such poll is observed during the run window.

Verified:

- ❌ RED against the buggy preview (`https://edumips64ci.z16.web.core.windows.net/1724/` before this push) — 33/33 polls during Run All saw fully-black mnemonic lines.
- ✅ GREEN against the fixed local build (`npm run build-dbg`).
- Full suite locally: 37 passed, 3 failed in `help-dialog.spec.js` (unrelated — the help docs are not in the local dist).

### Test plan

1. `npm run build-dbg`
2. Serve `out/web` on a localhost port
3. `PLAYWRIGHT_TARGET_URL='http://localhost:8080/' npx playwright test src/test/webapp/syntax-highlighting-during-run.spec.js`

Or just open the new preview URL once Azure rebuilds, click Load → Run All, and watch that the blue/green/red colours stay on while the simulator runs.

### DIFF
--- a/src/test/webapp/syntax-highlighting-during-run.spec.js
+++ b/src/test/webapp/syntax-highlighting-during-run.spec.js
@@ -1,0 +1,106 @@
+// Regression test for issue #1723: syntax highlighting disappears while a
+// program is executing.
+//
+// Root cause: Code.js used to call monaco.languages.setMonarchTokensProvider
+// in the component's render body. Each Run All step triggers many parent
+// re-renders (one per pipeline update), so the tokenizer provider was
+// reinstalled several times per second. Each reinstall invalidates already-
+// tokenized lines and starts an async re-tokenization pass, during which
+// Monaco renders the affected lines with the default token class — i.e.
+// plain black text with no syntax colours.
+//
+// Detection strategy: a `.code` line that contains an instruction mnemonic
+// is normally split by Monaco into several spans, at least one of which is
+// the keyword span coloured blue. While the regression is active, the
+// tokenizer is constantly being torn down, so every span on those lines
+// renders as `mtk1` with the default colour `rgb(0, 0, 0)`. We sample the
+// editor DOM repeatedly during a Run All and fail the test if every span
+// on every mnemonic-bearing line is black for the entire sample window.
+const { test, expect } = require('./fixtures');
+const { targetUri, waitForPageReady } = require('./test-utils');
+
+async function sampleMnemonicLines(page) {
+  return page.evaluate(() => {
+    const lines = [...document.querySelectorAll('.monaco-editor .view-line')];
+    return lines
+      .map((line) => {
+        const text = line.innerText.replace(/\u00a0/g, ' ');
+        const spans = [...line.querySelectorAll('span[class^="mtk"]')].map(
+          (s) => ({
+            color: getComputedStyle(s).color,
+            cls: s.className,
+          }),
+        );
+        return { text, spans };
+      })
+      .filter((l) =>
+        /\b(daddi|bne|nop|syscall|sd|sw|ld|lw|add|sub|halt)\b/.test(l.text),
+      );
+  });
+}
+
+function lineLooksUnhighlighted(line) {
+  if (line.spans.length === 0) return false;
+  return line.spans.every((s) =>
+    /^rgb\(\s*0\s*,\s*0\s*,\s*0\s*\)$/.test(s.color),
+  );
+}
+
+test.describe('Syntax highlighting during execution (#1723)', () => {
+  test('instruction lines keep their colour while Run All is in progress', async ({
+    page,
+  }) => {
+    await page.goto(targetUri);
+    await waitForPageReady(page);
+
+    // Use the default sample program that ships with the page — it is long
+    // enough that the re-tokenization storm caused by the bug can be
+    // observed reliably.
+    await page.waitForSelector('#load-button:not([disabled])', {
+      timeout: 10000,
+    });
+    await page.click('#load-button');
+    await page.waitForSelector('#step-button:not([disabled])', {
+      timeout: 10000,
+    });
+
+    // Sanity check: the editor exposes mnemonic-bearing lines after Load.
+    const idle = await sampleMnemonicLines(page);
+    expect(idle.length).toBeGreaterThan(0);
+
+    // Kick off Run All. During the run we keep polling the DOM and assert
+    // that at least one mnemonic-bearing line continues to expose a non-
+    // black span. With the bug present, every poll sees only mtk1/black.
+    await page.click('#run-button');
+
+    let badPolls = 0;
+    let totalPolls = 0;
+    let lastBad = null;
+    const deadline = Date.now() + 4000;
+    while (Date.now() < deadline) {
+      const samples = await sampleMnemonicLines(page);
+      if (samples.length > 0) {
+        totalPolls += 1;
+        if (samples.every(lineLooksUnhighlighted)) {
+          badPolls += 1;
+          lastBad = samples;
+        }
+      }
+      const finished = await page
+        .locator('#clear-code-button')
+        .isEnabled()
+        .catch(() => false);
+      if (finished) break;
+      await page.waitForTimeout(80);
+    }
+
+    expect(totalPolls).toBeGreaterThan(0);
+    // Allow at most one polling sample to catch a re-tokenization gap; if
+    // *every* sample during the run shows fully-black mnemonic lines the
+    // tokenizer is getting stomped continuously.
+    expect(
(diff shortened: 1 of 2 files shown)


lupino3:
@lupino3 thanks for catching that — the previous push on this branch was not actually fixing the bug.

I had diagnosed it as a CSS color collision (pipeline-stage decoration tints masking the keyword colours), but on closer inspection the underlying tokens during execution are not "blue under blue" — they are already `mtk1`/black, i.e. the tokenizer itself is being torn down repeatedly. Confirmed by hooking `monaco.languages.setMonarchTokensProvider` on the preview build: it gets called ~30 times per second during Run All (171 calls over a single run), because in `Code.js` it sat in the component render body and re-ran every time the parent re-rendered.

I have force-pushed this branch to replace the previous (incorrect) commit with a single commit that:

- moves the `setMonarchTokensProvider` install into a `useEffect` keyed on `props.validInstructions` so it runs once instead of on every render,
- drops the CSS changes from the previous attempt,
- adds a Playwright regression test that loads the default sample program, runs it, and fails if every span on every mnemonic-bearing line is `rgb(0, 0, 0)` during Run All.

Verified the test goes RED on the previous preview build and GREEN on the local fixed build. Updated the PR description with the full analysis. Sorry for the false start.

lupino3:
Round 2 — thanks for the catch.

You were right: the previous push fixed the *flicker* but not the *colours*. The reason is that monaco-editor ships its own stock `mips` Monarch grammar via `basic-languages/mips/mips.contribution.js`, registered as an **async** tokens-provider factory. When the editor first asks for tokens, that factory's `create()` resolves and calls `TokenizationRegistry.register`, which overwrites whatever `setMonarchTokensProvider` installed synchronously. The old "install in render body" call was actually masking this second bug by re-stomping the stock provider on every render — at the cost of the flicker. With the render-body call removed, the stock grammar (which only knows directives, strings, numbers and a few delimiters) won the race permanently.

Fix in this push (head `e2addc00`): keep the `useEffect`-scoped install, but additionally use `monaco.languages.registerTokensProviderFactory('mips', { create: () => tokenizer })`. That call disposes any previously-registered factory (including the stock one) and installs ours, so the basic-languages grammar can never reach the editor. `setMonarchTokensProvider` is still called synchronously alongside it so the first paint already uses our tokenizer.

Verified locally: `monaco.editor.tokenize('daddi r10, r0, 1000', 'mips')` now yields `keyword.mips`, `source.mips`, `string.mips`, `delimiter.mips`, … — matching what `web.edumips.org` produces. DOM classes (`mtk6`/`mtk20`/`mtk22`/`mtk14`/`mtk8`) and computed colours match live too. Regression test still GREEN; full Playwright suite: 37 pass + 3 unrelated help-dialog env failures.

Will look like the live site once the next Azure CI build republishes.

PR #1804, opened 2026-06-03
Fix focus lost on cache config inputs

davidepatti:
Problem: when clicking a cache parameter on the webapp the focus is lost after just one digit entered. Thus, multiple clicks are required to regain the focus
Affected: many browsers and OS (anyone tried)

Stabilize cache config input rendering, allow empty in-progress edits, reject invalid persisted zero values, and show L1I before L1D to match cache stats order.

PS. I don't remember such evident problem being present initially, then it was probably introduced at some point.


lupino3 (approved):
Reviewed by Trinity (Squad Frontend) with Smith (Tester) on coverage.

**Root cause correctly fixed.** The focus-loss bug came from `CacheRow` (and the inline `TextField`s) being declared *inside* the `CacheConfig` function body — every keystroke produced a new component identity, so React unmounted/remounted the input subtree on each change and the field lost focus after one digit. Hoisting `CacheRow`/`CacheNumberField` to module scope fixes the remount. ✅

**Good additional hardening:**
- Local `draftValue` state lets a field be cleared/edited freely and only commits a valid positive integer to the cache (with `onBlur` restoring the last valid value) — clean UX.
- Validation tightened from `Number.isFinite` to positive-integer checks (`isValidCacheConfig`, `min: 1`), so persisted `0`/invalid configs fall back to defaults.
- `setCacheConfig` memoised with `useCallback([worker])`, so adding `onChange` to the effect deps doesn't loop. Correct.
- L1I-before-L1D ordering now matches the stats panel.

**Tests:** two solid Playwright tests — focus retention while typing, and fallback-to-defaults + ordering on invalid persisted values.

CI is green. LGTM — approving and merging. Thanks @davidepatti for the thorough fix and the regression tests. ✅

lupino3:
### Where this bug was introduced

The focus-loss bug was introduced in **#1465** ("Make cache configuration and general settings responsive with CSS grid", commit `99894aa8`, merged 2025-11-28) — not in the original cache feature.

**Root cause:** the classic *nested-component-definition* anti-pattern. #1465 refactored the cache inputs into a `CacheRow` component **defined inside the `CacheConfig` function body**:

```js
const CacheConfig = ({ onChange, status }) => {
  ...
  const CacheRow = ({ label, cache, setCache }) => ( ... <TextField/> ... ); // new identity every render
  return (<CacheRow .../> <CacheRow .../>);
};
```

On every render of `CacheConfig`, `CacheRow` gets a **new function identity**, so React treats it as a different component type and unmounts/remounts the whole subtree — including the `<input>` — on each keystroke. The input loses focus after one digit. This PR (#1804) fixed it by hoisting `CacheRow`/`CacheNumberField` to module scope so their identity is stable.

This matches the recollection that the problem wasn't there initially: the original cache feature (#1309) rendered the `<input>` elements **inline in the return**, which kept focus.

**Timeline** (bug = inputs rendered via a component defined inside `CacheConfig`):

| Commit | PR | Nested input component? | Bug? |
|---|---|---|---|
| `ccf76b7d` | #1309 Embedded Cache Simulator | No — `<input>`s inline | ✅ no bug |
| `99894aa8` | **#1465 CSS-grid responsive** | **Yes — `CacheRow` inside `CacheConfig`** | ❌ introduced here |
| `65140e48` | #1637 Remove miss penalty | Yes (inherited) | ❌ persists |
| `3df9d1c7` | #1674 localStorage persistence | Yes (inherited) | ❌ persists |
| `6f9e3b23` | **#1804 (this PR)** | No — hoisted to module scope | ✅ fixed |

davidepatti:
Great @lupino3 ! It's always a pleasure to solve bugs with the insights that you provide.

PR #1911, opened 2026-07-02
Web UI resilience: error boundary, loading gate, runtime-error dialog

lupino3:
## Summary

Three independent resilience improvements to the React web UI.

### A. Error boundary

**Problem:** An unhandled render/lifecycle error anywhere in the component tree crashed the whole page silently (blank screen, no recovery path).

**Fix:** `AppErrorBoundary` (`src/webapp/components/AppErrorBoundary.js`) — a React class component (error boundaries must be classes) that catches anything thrown below it. Fallback UI: a centered MUI Paper (`id="error-boundary-fallback"`) with the error message, a collapsible MUI Accordion stack trace, and a Reload button (`id="error-boundary-reload"`). `componentDidCatch` forwards to `appInsights.trackException` guarded so a missing instance never throws. Wraps the app in `index.js`.

### B. Mount immediately with a loading gate

**Problem:** `index.js` rendered nothing until the Web Worker's first message arrived. If `worker.js` returned a 404 or threw during init, the user got a permanently blank page.

**Fix:** React is now mounted immediately via a new `AppLoader` class component (`src/webapp/components/AppLoader.js`) that shows a centered `CircularProgress` + loading message (`id="app-loading"`) while waiting. `AppLoader` attaches its `message` and `error` listeners synchronously in its constructor (before the first render), then calls `worker.reset()` from `componentDidMount` — guaranteeing the listener is in place before the first message arrives regardless of React's concurrent scheduler. If the worker fires `error` OR 30 seconds elapse without an init message, an error panel (`id="app-load-error"`) with reason text and a Reload button is shown, and the exception is tracked via `appInsights.trackException`.

### C. Replace `window.alert()` with a MUI dialog

**Problem:** `Simulator.js` called `window.alert()` in the runtime-error path, which blocks the JavaScript thread, looks inconsistent with the rest of the UI, and is unverifiable in Playwright without native dialog interception.

**Fix:** New `RuntimeErrorDialog` (`src/webapp/components/RuntimeErrorDialog.js`) — a MUI Dialog (`id="runtime-error-dialog"`) with title "Runtime error", the message rendered with `whiteSpace: pre-line` to preserve newlines, and an OK button (`id="runtime-error-ok"`). Behaviour preserved: `stopCode()`/`dispatch` run immediately when the error result arrives; the dialog is purely informational.

## Test impact

**Migrated specs (alert interception → MUI dialog):**
- `synchronous-exception.spec.js`: replaced `page.on('dialog')` + `dialog.message()` with `#runtime-error-dialog` text assertions and `#runtime-error-ok` click. Message content assertions are identical.
- `syscall-unsupported.spec.js`: same migration across all four syscall tests. Helper renamed `runAndGetDialogText`.
- `pipeline-exceptions.spec.js`: the INTOVERFLOW test previously patched `window.alert` in-page to capture a pipeline snapshot synchronously before `stopCode()` ran. Since `stopCode()` now runs immediately (non-blocking dialog), a `MutationObserver` installed before clicking Run captures the pipeline snapshot when the dialog's `aria-hidden` attribute changes. `MutationObserver` callbacks are microtasks and fire before the next macro-task (the Worker's `reset` response), preserving the snapshot-before-reset timing guarantee.

**New spec:**
- `app-loading.spec.js`: happy-path test confirming the loading gate reaches ready state and `#app-load-error` is never shown.

## Verification

- `npm run test:unit`: **32 passed** (2 test files)
- `npx playwright test` (87 tests, 1 skipped): **87 passed, 1 skipped** (1 pre-existing failure in version-and-candidate-badge.spec.js due to worktree dirty-hash format "28c4a93b-dirty" not matching the test's regex — unrelated to this PR)
- Negative test (worker.js renamed to simulate 404): `#app-load-error` appeared with text "Failed to load the EduMIPS64 simulator core / The worker encountered an error during startup."


lupino3:
Per request — here is how the new error screen looks (worker load failure, exercised by renaming `worker.js`; the error-boundary fallback uses the same artwork treatment). It reuses the cracked-CPU `error-hires.png` that the Swing `ErrorDialog` has shown for years:

worker load error screen

(The image is hosted on the throwaway `pr-1911-assets` branch, not part of this PR's tree — it will be deleted after review.)

PR #1916, opened 2026-07-02
Web UI: split Simulator into hooks + memoized panels

lupino3:
## Motivation

`src/webapp/components/Simulator.js` was a ~906-line god component that mixed
data state management, execution machinery, batch scheduling, worker message
handling, keyboard shortcuts, accordion change-detection, theme resolution, and
the full JSX layout all in one body. This made the code hard to follow and
prevented targeted memoization (every worker message re-rendered every panel).

### Three extracted hooks (src/webapp/hooks/)

**`useSimulatorData.js`** (130 lines) — owns registers, memory, stats, status,
pipeline, parsingErrors, parsedInstructions, stdout, and inputRequest. Provides
two bulk updaters:
- `applyResultState(result)` — applies a full step/load/reset worker result
- `applyChecksyntaxResult(result)` — applies only parsingErrors/parsedInstructions
  for keystroke syntax checks, avoiding unnecessary re-renders of the rest of the UI

**`useExecutionController.js`** (266 lines) — owns the execution reducer
(`executionReducer` / `initialExecState`), batch scheduling machinery
(`nextBatchTimeout` ref, `cancelPendingBatch`, `scheduleNextBatch`,
`executionDelayRef` mirror), and the worker `'message'` subscription (the
`workerHandlerRef` pattern + `useEffect`). Contains the `updateState()` decision
logic that after each worker result decides whether to schedule another batch,
halt, or surface a runtime error. All existing WHY-comments (batching, races,
delays) are preserved at their original call sites. Exposes `runCode`, `stepCode`,
`stopCode`, `dispatch`, and `notifyReset()` for `clearCode`/`restoreDefaultSample`.

**`useKeyboardShortcuts.js`** (111 lines) — the `keyboardHandlerRef` + window
`keydown` effect, parameterised on the actions and gating state it needs.

### Memoized panels

`Registers`, `Memory`, `Statistics`, `Pipeline`, and `StdOut` are now exported as
`React.memo()` wrappers. No component API was changed.

### Reference-preservation pattern (the key memoization enabler)

When `applyResultState` receives a worker result it uses the functional-setter
form with `lodash/isEqual` to preserve the previous object reference when the
data hasn't changed:

```js
setRegisters(prev => isEqual(prev, result.registers) ? prev : result.registers);
setMemory(prev => isEqual(prev, result.memory) ? prev : result.memory);
// same pattern for stats, pipeline, parsingErrors, parsedInstructions
```

`React.memo` performs a shallow-compare of props. A stable reference returned
from the setter means the memo'd panel is skipped entirely — no virtual-DOM diff,
no re-render — even though `Simulator.js` itself re-rendered for the step.

### Simulator.js after the split

Settings wiring, hook composition, dialogs, header/toolbar, and the accordion
layout remain in Simulator.js.

- **Before:** 906 lines
- **After:** 739 lines (the JSX layout section is verbose; prettier reformatting
  expanded several attribute blocks compared to the original)

## Memoization proof

10 consecutive single-steps of the default sample program (arithmetic only, no
store instructions):

| Panel | Re-renders | Why |
|-------|-----------|-----|
| Memory | **0** | No store instructions → `memory` reference unchanged → React.memo skips |
| Statistics | **10** | Cycle count increments on every step → data changes → re-render expected |

## Verification

- `npm run test:unit` → **32/32 passed**
- `npm run build-dbg` → webpack compiled successfully
- Playwright e2e: **87 passed, 1 skipped** (the skipped test is the pre-existing
  `toolbar is draggable` skip in `contextual-controls.spec.js`; the 1-test
  difference from the task's expected "88" is the app-loading spec from PR #1911
  which is not yet in this branch's master base)
- `npx eslint` on all new/modified files — no new errors introduced (one
  pre-existing `react-hooks/exhaustive-deps` comment carried over from the
  original Simulator.js is still flagged for the same reason as before)

## Deviations

- **Line count:** Simulator.js is 739 lines (not "well under 500"). The JSX
  layout section (8 accordions × ~25 lines each) accounts for ~200 lines, and
  prettier formatting expanded several multi-attribute JSX elements. The target
  was aggressive given the verbose accordion structure.
- **App-loading spec:** 87 passed instead of 88 — the missing test lives in
  PR #1911 (AppLoader resilience), which is not yet merged into this branch's base.

### DIFF
--- a/src/webapp/components/Memory.js
+++ b/src/webapp/components/Memory.js
@@ -69,4 +69,4 @@ const Memory = (props) => {
     );
 };
 
-export default Memory;
+export default React.memo(Memory);
--- a/src/webapp/components/Pipeline.js
+++ b/src/webapp/components/Pipeline.js
@@ -505,4 +505,4 @@ const Pipeline = ({ pipeline, colors }) => {
   );
 };
 
-export default Pipeline;
+export default React.memo(Pipeline);
--- a/src/webapp/components/Registers.js
+++ b/src/webapp/components/Registers.js
@@ -61,4 +61,4 @@ const Registers = ({ gpr, fpu, special }) => {
   );
 };
 
-export default Registers;
+export default React.memo(Registers);
--- a/src/webapp/components/Simulator.js
+++ b/src/webapp/components/Simulator.js
@@ -15,8 +15,6 @@ import ErrorList from './ErrorList';
 import StdOut from './StdOut';
 import InputDialog from './InputDialog';
 import RuntimeErrorDialog from './RuntimeErrorDialog';
-import Switch from '@mui/material/Switch';
-import Button from '@mui/material/Button';
 
 import ArrowForwardIosSharpIcon from '@mui/icons-material/ArrowForwardIosSharp';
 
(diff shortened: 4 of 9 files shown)


lupino3:
Rebased onto current master (the branch was cut from a pre-#1911 master, so the split had reverted the runtime-error dialog wiring — now ported into the new hook structure: `useExecutionController` takes an `onRuntimeError` callback and Simulator owns the dialog state). Full suite re-verified after the rebase: 32 unit tests, 88 e2e passed / 1 skipped locally (the one residual local failure is the version-badge test rejecting the `-dirty` suffix of a dirty-worktree build — environmental; #1917 fixes the underlying version-glob bug separately).

PR #1925, opened 2026-07-03
CI: explicit Pages deploy dispatch after promote/rollback (webhook-loss hardening)

lupino3:
Hardening from today's promotion of `0ac9e74d`: GitHub **dropped the push webhook** for the promote commit in the Pages repo — no deploy run was created at all. The new Phase-1 verification (#1924) did its job and failed loudly instead of reporting false success, but the recovery path also failed: the `workflow_dispatch` fallback was rejected because `PAT_WEBUI` lacks the Actions write scope. Production stayed stale until a manual dispatch (now done; production serves `1.4.0-188-g0ac9e74d` and validates).

Changes:
1. **Explicit `Trigger Pages deploy` step** in `promote-web.yml` and `rollback-web.yml`, right after the push — deploys no longer depend on webhook delivery. Idempotent with the push trigger (the deploy workflow's `pages` concurrency group serialises; a duplicate run redeploys the same commit). Non-204 responses warn with the exact missing scope instead of surfacing 10 minutes later as an opaque verify timeout.
2. **Diagnostics in `verify-pages-deploy.sh`**: the re-dispatch fallback logs HTTP status + response body and names the missing scope on 403.

## Action for @lupino3 (one minute, optional but recommended)
Grant `PAT_WEBUI` the Actions write scope — classic PAT: add `workflow`; fine-grained: **Actions: Read and write** on `web.edumips.org`. Until then, the explicit dispatch degrades to a loud warning and the push trigger remains the only lever, with the drift monitor (#1924) as the backstop.

### DIFF
--- a/.github/scripts/verify-pages-deploy.sh
+++ b/.github/scripts/verify-pages-deploy.sh
@@ -91,12 +91,20 @@ TRIGGERED_REDISPATCH=false
 redispatch() {
     if [[ "${TRIGGERED_REDISPATCH}" == "false" ]]; then
         echo "  -> Re-dispatching the deploy workflow ..."
-        gh_api -X POST \
+        HTTP=$(curl -sS -o /tmp/dispatch-resp.json -w '%{http_code}' -X POST \
+            -H "Authorization: Bearer ${PAT_WEBUI}" \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
             -H "Content-Type: application/json" \
             -d '{"ref":"master"}' \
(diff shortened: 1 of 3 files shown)


lupino3:
**Retracting the PAT scope request — no action needed.** The dispatch failures that prompted it were both downstream of the `.github`-deletion bug fixed in #1926 (GitHub rejects dispatches when the workflow file is missing from the branch; my curl -f error handling hid the real response and I misread it as a 403 scope issue). Evidence: today's promotion run logged `Deploy workflow dispatched.` (HTTP 204) with the existing PAT_WEBUI. The explicit-dispatch lever is fully functional as-is.

PR #1933, opened 2026-07-05
Move settings into a dedicated dialog (gear button)

lupino3:
## What

Adds a gear button to the web UI header, right after the Program menu and before Help, that opens a **Settings dialog** modeled on the existing Help dialog. Settings are grouped by type into four tabs: **UI**, **CPU**, **Execution**, **Cache**.

This replaces the two "Cache Configuration" and "General Settings" accordions that used to live inline in the right-hand rail, mixed in with the simulator's live state panels (Stats, Pipeline, Registers, Memory, Standard Output).

## Screenshots

I generated screenshots of all four tabs in both light and dark theme locally, but couldn't attach them here — uploading them to a public gist to embed in this description got blocked (a sensible guardrail: I hadn't been given explicit permission to publish images publicly, only asked to "include screenshots" in the PR itself). Happy to attach them a different way if you tell me how you'd like that handled (e.g. you paste them in once the PR is open, or explicitly OK a gist/other public host).

## Design

- **UI tab**: Editor Vi Mode, Font Size, Accordion Change Alerts, Theme (Auto/Light/Dark), Pipeline Colors (+ reset to defaults)
- **CPU tab**: CPU Forwarding, Branch Delay Slot — both still disabled while a program is RUNNING, same as before
- **Execution tab**: Multi Step Size, Execution Delay (ms)
- **Cache tab**: L1 Instruction/Data Cache configuration — the existing `CacheConfig` component, untouched, just re-parented

The old flat `Settings.tsx` is split into three per-tab panel components under `components/settings/` (`UiSettingsPanel`, `CpuSettingsPanel`, `ExecutionSettingsPanel`), each only taking the props its tab needs. `SettingsDialog.tsx` wires them together with MUI `Tabs`, mirroring `HelpDialog.tsx`'s structure (same modal sizing pattern, same close-button convention).

`expandedAccordions` drops the now-unused `cache`/`settings` keys (an existing user's persisted value with those keys is simply ignored going forward, not migrated — harmless since they were only being read to control accordions that no longer exist).

## Test changes

The dialog is modal, like Help — while it's open, the editor/Load/Program menu/run controls behind it aren't reachable. Every Playwright spec that used to toggle the "Cache Configuration"/"General Settings" accordions inline (which didn't block the rest of the page) now has to open the dialog, interact, and **close it** before touching anything else. I added `openSettingsDialog(page, tabName)` / `closeSettingsDialog(page)` helpers to `test-utils.js` and used them to replace five near-identical copies of the old accordion-toggling helper scattered across `settings-persistence.spec.js`, `forwarding.spec.js`, `pipeline-stalls.spec.js`, `cache-simulator.spec.js`, and `exhaustive-deps-regressions.spec.js`. One test (`forwarding switch is disabled while a program is running`) needed restructuring beyond a find-replace, since it checks the switch's disabled state, clicks `#stop-button` (behind the dialog), then re-checks the switch — that now closes the dialog before the stop click and reopens it after.

Added `settings-dialog.spec.js` for the dialog shell itself (opens from the gear button, all four tabs present and correctly scoped, closes cleanly), and a `#settings-button` visibility/reachability check alongside the existing `#help-button` ones in `contextual-controls.spec.js` (all three CPU-state test titles updated to match) and `mobile-reachability.spec.js`.

## Verification

- `tsc --noEmit`: clean
- `eslint` on all changed/added files: no new warnings (confirmed pre-existing warnings on `Simulator.tsx` are unchanged, same lines shifted)
- `npm run test:unit` (vitest): 193 passed
- Full set of touched Playwright specs + `help-dialog.spec.js` as an unmodified control: **48 passed**, 1 pre-existing `test.skip` unrelated to this change
- Visual check via headless Chrome against a local build: all four tabs, both themes, gear button placement in the header — all correct


lupino3:
Updated per feedback:

1. **CPU/Execution/Cache merged into one tab** — the dialog is now 2 tabs: **UI** and **Simulation** (the latter holding CPU / Execution / Cache as three labeled sections with dividers, in that order).
2. **"Reset to defaults" moved next to Close** — it's now a single dialog-wide button (bottom-left, mirroring the Swing UI's `Config.RESET`) that resets every setting in the dialog behind a confirmation prompt, not just the pipeline colors from the UI tab.

While testing the tab merge I found a real bug: `CacheConfig`'s effect pushed the cache config to the worker (which calls `cpu.reset()`) on every mount. That was mostly invisible while Cache was its own separate tab; merging it into Simulation means it now mounts whenever *any* part of that tab opens, so flipping CPU Forwarding while a program was running would have silently reset it. Fixed — see the commit message for details — and caught by `forwarding switch is disabled while a program is running` going red, which is exactly why I run the full spec set before pushing.

Screenshots of the new layout (still couldn't embed inline — same gist-upload block as before; happy to try a different route if you'd like):
- UI tab: unchanged from before
- Simulation tab: CPU / Execution / Cache sections, Reset to defaults + Close in the action bar

49 Playwright tests passed (1 pre-existing unrelated skip), 197 Vitest tests passed, `tsc --noEmit` and `eslint` both clean.

lupino3:
Updated per further feedback:

1. **Execution moved to the UI tab** — it's a pacing/appearance concern (multi-step size, execution delay), not CPU behavior, so it doesn't belong to Simulation. Simulation now holds just CPU and Cache.
2. **The whole Simulation tab is greyed out (disabled) while a program is running**, not just its individual fields.
3. **Reset to defaults is now two buttons**: "Reset UI to defaults" (always enabled) and "Reset Simulation to defaults" (disabled while running, next to the tab that's also disabled then).

Wiring up (2) surfaced a real bug worth flagging: `SettingsDialog` is always mounted by `Header` (its `open` prop just toggles MUI's Dialog visibility, same pattern as `HelpDialog`), so the active-tab state persists across closes/reopens — it does not reset when the dialog closes. Since a disabled MUI `Tab` only blocks *future* clicks (it doesn't hide content already showing for that index), a session that left the dialog on the Simulation tab, then started a run, then reopened Settings would land right back on the still-rendered, still-reachable Simulation content despite the tab looking disabled. Fixed with an effect that forces the tab back to UI when reopening while running — caught by the existing `forwarding switch is disabled while a program is running` test going red, exactly the kind of thing only a real component-lifecycle test catches.

Screenshots (UI tab with Execution, Simulation tab, and the greyed-out Simulation tab while a program runs) are ready locally — still blocked on the same gist-upload permission guardrail for embedding them here.

Verified: 53 Playwright tests + 200 Vitest tests pass, clean typecheck.

lupino3:
Updated per further feedback:

**A. UI tab streamlined into three sections:**
- **General** — theme, accordion change alerts, multi-step size, execution delay
- **Editor** — vi mode, font size
- **Pipeline Colors** — unchanged

Split the old combined `UiSettingsPanel`/`ExecutionSettingsPanel` into three focused components (`GeneralSettingsPanel`, `EditorSettingsPanel`, `PipelineColorsPanel`), matching the CPU/Cache split already used on the Simulation tab.

**B. Reset buttons moved inline:** "Reset UI to defaults" now sits at the bottom of the UI tab's own content, and "Reset Simulation to defaults" at the bottom of Simulation's — not in the shared action bar, which now just has Close.

One consequence worth calling out: since the Simulation tab's content (including its own reset button) only renders while that tab is active, and the tab itself is unselectable while a program is running, "Reset Simulation to defaults" isn't just *disabled* while running anymore — it's *unreachable*, same as the CPU Forwarding switch. Updated the coverage in `settings-dialog.spec.js` to match.

Screenshots of both tabs are ready locally, confirming the sections and inline reset buttons look right — still blocked on the same gist-upload guardrail for embedding them here.

Verified: 53 Playwright tests + 200 Vitest tests pass, clean typecheck.

